PR #82, opened 2016-11-11
Limit queue size

maksimf:


lib/cache/adapters/memory_adapters/queue_adapter.rb
```diff
@@ -7,18 +7,27 @@ class QueueAdapter
           def initialize(size)
             @size = size
```
patricioe: @maksimf can we make a tiny name refactor here? I would name this: @max_size and rename @size_counter to @current_size ?

lib/cache/adapters/memory_adapters/queue_adapter.rb
```diff
@@ -7,18 +7,27 @@ class QueueAdapter
...
 
           # Get all items from the queue
           def clear
             items = []
 
-            loop { items << @queue.pop(true) }
+            loop do
+              items << @queue.pop(true)
+
+              @size_counter.decrement
```
patricioe: Why not moving this out of the loop and set it to zero when it's done?

spec/cache/adapters/memory/queue_adapter_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  it 'throws exception if queue size is reached' do
+    expect { adapter.add_to_queue('foo') }.to raise_error(ThreadError)
+  end
+
+  it 'sets correct size_counter after clear' do
+    adapter.clear
+
+    expect(adapter.instance_variable_get(:@size_counter).value).to eq(0)
+  end
```
patricioe: Could you add one more test that add say 10 impressions, and check the size is 10?

maksimf:
@patricioe All done

lib/cache/adapters/memory_adapters/queue_adapter.rb
```diff
@@ -4,24 +4,34 @@ module Adapters
...
 
-            loop { items << @queue.pop(true) }
+            loop do
+              items << @queue.pop(true)
+
+            end
 
           rescue ThreadError
             # Last queue item reached
+            @current_size.value = 0
```
patricioe: does it always reach this point?

lib/cache/adapters/memory_adapters/queue_adapter.rb
```diff
@@ -4,24 +4,34 @@ module Adapters
...
           end
 
           # Get all items from the queue
           def clear
             items = []
 
-            loop { items << @queue.pop(true) }
+            loop do
+              items << @queue.pop(true)
+
```
patricioe: ok, I think I made a mistake.

I think it CAN happen and will happen that producer will continue adding elements to the queue while this is taking things out right?
If so, I think it's safe to say that we should decrement here every time you take one out.

I'm afraid that 
```ruby
rescue ThreadError
   # Last queue item reached
   # another element is added right between the catch clause and the next time, due to the nature of concurrency
  # and we set by mistake the size to zero, when actually there was one or more elements. 
  @current_size.value = 0

  items
end
```

maksimf:
@patricioe Fixed

lib/cache/adapters/memory_adapters/queue_adapter.rb
patricioe: how do we prevent race conditions here?

maksimf: @patricioe I've removed that line

lib/cache/adapters/memory_adapters/queue_adapter.rb
```diff
@@ -4,24 +4,32 @@ module Adapters
...
-            @size = size
+          def initialize(max_size)
+            @max_size = max_size
             @queue = Queue.new
+            @current_size = Concurrent::AtomicFixnum.new(0)
           end
 
           # Adds data to queue in non-blocking mode
           def add_to_queue(data)
+            fail ThreadError if @current_size.value >= @max_size
```
patricioe: queue_adapter.rb uses it and does not catch `ThreadError`.

Is that an outdated code?
```ruby
          # Adds data to queue in non-blocking mode
          def add_to_queue(data)
            @queue.push(data)
          end
```

PR #84, opened 2016-11-14
Refactor impressions

maksimf:


lib/cache/senders/impressions_sender.rb
```diff
@@ -0,0 +1,59 @@
...
+            Thread.new do
+              loop do
+                post_impressions
+
+                sleep(randomize_interval(@config.impressions_refresh_rate))
+              end
+            end
+          end
+
+          @config.logger.info('Started impressions service') unless ENV['SPLITCLIENT_ENV'] == 'test'
```
patricioe: should we move ` @config.logger.info('Started impressions service')` between:
```ruby
Thread.new do
  @config.logger.info('Started impressions service') # <----- ?
  loop do
     ....
```

??

lib/cache/senders/impressions_sender.rb
```diff
@@ -0,0 +1,59 @@
...
+      class ImpressionsSender
+        def initialize(impressions_repository, config, api_key)
+          @impressions_repository = impressions_repository
+          @config = config
+          @api_key = api_key
+        end
+
+        def call
+          # Disable impressions if @config.impressions_queue_size == -1
+          if @config.impressions_queue_size < 0
```
patricioe: do we have a unit test that post impressions when it is disabled to make sure it doesn't blow?

maksimf:
@patricioe Done, fixed

spec/engine_spec.rb
```diff
@@ -272,6 +272,18 @@
...
+
+      context 'when impressions are disabled' do
+        let(:config) do
+          { logger: Logger.new('/dev/null'), cache_adapter: cache_adapter, impressions_queue_size: -1 }
+        end
+
+        it 'works when impressions are disabled' do
+          expect(subject.get_treatments('21', ["sample_feature", "beta_feature"])).to eq(
+            { sample_feature: 'off', beta_feature: 'off' }
+          )
```
patricioe: shouldn't we check that there are no impressions in the queue here?

lib/splitclient-rb/version.rb
```diff
@@ -1,3 +1,3 @@
 module SplitIoClient
-  VERSION = '3.1.0-rc12'
+  VERSION = '3.2.0'
```
patricioe: Please don't change this here yet.

lib/engine/api/client.rb
```diff
@@ -24,6 +37,15 @@ def api_client
           builder.adapter :net_http_persistent
         end
       end
+
+      def common_headers(api_key, config)
+        {
+          'Authorization' => "Bearer #{api_key}",
+          'SplitSDKVersion' => SplitIoClient::SplitFactory.sdk_version,
+          'SplitSDKMachineName' => config.machine_name,
+          'SplitSDKMachineIP' => config.machine_ip
```
patricioe: do we need to cherry pick the referer commit here?

lib/cache/senders/impressions_sender.rb
```diff
@@ -0,0 +1,59 @@
...
+          if ENV['SPLITCLIENT_ENV'] == 'test'
+            post_impressions
+          else
+            Thread.new do
+              @config.logger.info('Starting impressions service') unless ENV['SPLITCLIENT_ENV'] == 'test'
+
+              loop do
+                post_impressions
+
+                sleep(randomize_interval(@config.impressions_refresh_rate))
```
patricioe: do we need to cherry pick or rebase devleopment to make sure your Utilities class is used instead?

PR #87, opened 2016-11-16
Move localhost behavior to localhost_split_factory

maksimf:


patricioe:
@maksimf mind:
 * providing  a snippet on how to build both, localhost and regular? Is it using the only one Split_factory_builder? this one should bind one or another implementation based on the API key used
 * why both factories don't have a common class called `factory` ?
 * Please update the detailed-Readme.md
 * Add unit test that calls split_factory_builder with `localhost` key and asserts that the factory type returned is `locahost_split_factory`
 * Add unit test that calls split_factory_builder with `some_random_key` key and asserts that the factory type returned is the real `split_factory`

maksimf:
@patricioe 

1) Examples provided in the updated detailed-Readme, see "Localhost mode" section
2) I've unified creation of the desired factory, now you need to run `SplitIoClient::SplitFactoryBuilder.build('localhost or API_KEY')`
3) Unit tests added

maksimf:
^ Something strange with build on CI. Looks like rspec is not installed there.

Detailed-README.md
```diff
@@ -58,12 +58,41 @@ For advance use cases you can also obtain a `manager` instance from the factory.
...
+reporting_v2 on
+double_writes_to_cassandra off
+new-navigation v3
+
+```
+
+To use SDK in the localhost mode you should pass `localhost` as an API key like this:
+
+```
+factory  = SplitIoClient::SplitFactoryBuilder.build('localhost').client
```
patricioe: @maksimf drop the `client` at the end since you are creating a factory.

Detailed-README.md
```diff
@@ -91,7 +120,7 @@ The following values can be customized:
 *default value* = `5`
 
 **features_refresh_rate** : The SDK polls Split servers for changes to feature roll-out plans. This parameter controls this polling period in seconds
-
+split_client.get_treatment('user_id','feature_name', attr: 'val')
```
patricioe: is this misplaced here?

PR #89, opened 2016-11-18
Pass options from cli

maksimf:


exe/splitio
```diff
@@ -31,7 +80,9 @@ begin
    exit(1)
  end
 
-config = YAML.load_file(options[:config_path])
-config.merge!(mode: :producer, cache_adapter: :redis)
+config = YAML.load_file(config_path)
+config
+  .merge!(mode: :producer, cache_adapter: :redis)
+  .merge!(options)
```
patricioe: does the param in command line take precedence over the file one?

It looks good. Mind adding docs?

maksimf: @patricioe Yep, it does. Docs added

PR #92, opened 2016-11-21
Add referer info

maksimf:


maksimf:
@patricioe I'm fetching client info from the machine's hostname. Is that right?

lib/engine/api/client.rb
```diff
@@ -43,9 +43,18 @@ def common_headers(api_key, config)
...
-          'SplitSDKMachineIP' => config.machine_ip
+          'SplitSDKMachineIP' => config.machine_ip,
+          'Referer' => referer
         }
       end
+
+      def referer
+        result = SplitIoClient::SplitFactory.sdk_version
+
+        result = "#{result}::#{SplitIoClient::SplitConfig.get_hostname}" unless SplitIoClient::SplitConfig.get_hostname == 'localhost'
```
patricioe: can this be nil?

Merging for now, but if the answer is yes, let's address it on development.

patricioe:
Output: `"POST /api/testImpressions/bulk HTTP/1.1" 200 - "RubyClientSDK-3.1.2::patriciaguesMBP.lan" "Ruby" 819`

PR #93, opened 2016-11-21
Fix passenger smart spawning mode issue

maksimf:


lib/cache/senders/impressions_sender.rb
```diff
@@ -18,20 +18,30 @@ def call
...
           else
-            Thread.new do
-              @config.logger.info('Starting impressions service')
+            impressions_thread
 
-              loop do
-                post_impressions
-
-                sleep(randomize_interval(@config.impressions_refresh_rate))
+            if defined?(PhusionPassenger)
```
patricioe: @maksimf should this code live in this repo?

Part of me think that either:
- We put this snippet in another repo for users of passenger, or
- Provide these lines in the docs for users to add them to their set up.

Would that work?

lib/cache/senders/impressions_sender.rb
```diff
@@ -18,20 +18,30 @@ def call
...
+              PhusionPassenger.on_event(:starting_worker_process) do |forked|
```
patricioe: @maksimf did you test this yourself? any other consequence of resurrenting the threads?

maksimf: I did test it, didn't notice any consequences

lib/cache/senders/impressions_sender.rb
maksimf: @patricioe I thought it over and maybe that's not the best idea to do so, because it would mean that we'd need to provide direct access to impressions/metrics threads to the users of the SDK and make them run those threads manually in case of Passenger, seems kind of breaking of encapsulation for me. I'd say it would be better to leave it this way

patricioe:
@maksimf could you describe to me how did you test this?

maksimf:
@patricioe Firstly, I've launched passenger in the standalone mode with at least 2 instances like so:

```bash
passenger start --spawn-method=smart --min-instances=2
```

Then, I made sure that the thread which clears impressions runs in process A, whereas impressions are stored in the memory of the B or C processes. (A would be AppPreloader, B, C - RubyWorkers).

After I've introduced that change I logged impressions at the moment they were cleared from the queue and everything looks ok now (all impressions are cleared from the queue). Now there are 3 threads in all processes (A, B, C).

dendril:
@nahuelrebollo please read comments from @maksimf if you have any question.

PR #94, opened 2016-11-22
Automatically refresh localhost splits

maksimf:


Detailed-README.md
```diff
@@ -76,13 +76,18 @@ new-navigation v3
...
 ```
 
-By default SDK will look in your home directory (i.e. `~`) for a `.split` file, but you can specify a different 
+By default SDK will look in your home directory (i.e. `~`) for a `.split` file, but you can specify a different
 file name (full path) to look for the file (note: you must provide absolute path):
 
+When in localhost mode you can make use of the SDK ability to automatically refresh splits from file, to do that just specify reload rate in seconds like this:
+
+```ruby
+factory = SplitIoClient::SplitFactoryBuilder.build('localhost', path: '/where/to-look-for/<file_name>', reload_rate: 3)
```
patricioe: @maksimf is there a concept of file watcher more of out of the box? for two reasons:
- to prevent having a thread and moaking the code a bit more complicated.
- to avoid adding extra configs to the factory method.

patricioe: Do we need a `{ path: '/where/to-look-for/<file_name>', reload_rate: 3) } `?

maksimf: Yes, that's an example on how to use watcher

PR #97, opened 2016-11-28
Lower number of requests to Redis

maksimf:


lib/splitclient-rb/split_factory.rb
```diff
@@ -97,9 +97,12 @@ def get_treatments(key, split_names, attributes = nil)
         bucketing_key, matching_key = keys_from_key(key)
         bucketing_key = matching_key if bucketing_key.nil?
 
-        treatments = @splits_repository.get_splits(split_names).each_with_object({}) do |(name, data), memo|
-          memo.merge!(name => get_treatment(key, name, attributes, data, false))
-        end
+        treatments =
+          if @config.cache_adapter == :redis
+            @splits_repository.adapter.pipelined { multiple_treatments(split_names, key, attributes) }
```
dendril: @maksimf we could abstract "pipelinig" concept so you don't need to add the if. Does it make sense?

maksimf:
@dendril It's ready

PR #98, opened 2016-11-28
Send bucketing keys to impressions

maksimf:


dendril:
@maksimf I have checked the services still returns 200 even if the API doesn't handle the field and looks to be working.

cc @sarrubia

PR #101, opened 2016-12-01
New Redis namespace

maksimf:


maksimf:
@patricioe This PR is ready

lib/splitclient-rb/split_config.rb
```diff
@@ -249,6 +251,10 @@ def self.default_redis_url
       'redis://127.0.0.1:6379/0'
     end
 
+    def self.default_redis_namespace
+      "SPLITIO/ruby-#{VERSION}"
```
patricioe: I can't find where VERSION is coming from.

maksimf: This constant comes from the class in the current namespace (i.e. `SplitIoClient::VERSION`) which is defined here: `lib/splitclient-rb/version.rb`

patricioe:
there are conflicts to merge @maksimf

maksimf:
@patricioe resolved

PR #106, opened 2016-12-22
Log all exceptions when calling API

maksimf:
I'm not 100% this will fix currently failing tests, but this is definitely the right thing to do anyway

### DIFF
--- a/lib/engine/api/client.rb
+++ b/lib/engine/api/client.rb
@@ -34,7 +34,7 @@ module SplitIoClient
           end
         end
       rescue StandardError => e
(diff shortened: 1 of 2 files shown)


lib/engine/api/client.rb
```diff
@@ -14,6 +14,8 @@ def get_api(url, config, api_key, params = {})
 
           config.logger.debug("GET #{url}") if config.debug_enabled
         end
+      rescue StandardError => e
+        Logger.new('log/api_client.log').info("#{e}\nURL:#{url}\nparams:#{params}")
```
patricioe: @maksimf can we use the already set up logger instead of a new file? So customer know where to expect it.

PR #107, opened 2016-12-23
Rename block_until_ready to ready

maksimf:


maksimf:
@patricioe this PR can be merged

lib/splitclient-rb/split_factory.rb
```diff
@@ -22,7 +22,7 @@ def initialize(api_key, config_hash = {})
       @client = SplitClient.new(@api_key, @config, @adapter, @splits_repository, @segments_repository, @impressions_repository, @metrics_repository)
       @manager = SplitManager.new(@api_key, @config, @adapter, @splits_repository)
 
-      @sdk_blocker.block if @config.block_until_ready
+      @sdk_blocker.block if @config.block_until_ready > 0
```
patricioe: I'm missing the relation between config.ready and config.block_until_ready.

Shouldn't you get the value of @config.ready and use it here and also set the value of @config.block_until_ready to @config.ready is the former exists for backward compatibility?

maksimf: @patricioe the key is here: https://github.com/splitio/ruby-client/blob/refactor/rename_block_until_ready/lib/splitclient-rb/split_config.rb#L55

You can specify either `ready` or `block_until_ready`, but internaly I use `block_until_ready` variable

PR #109, opened 2016-12-27
Fix threads for passenger

maksimf:


lib/cache/repositories/splits_repository.rb
```diff
@@ -17,6 +17,8 @@ def initialize(adapter, config)
         end
 
         def add_split(split)
+          return unless split[:name]
```
patricioe: so you don't add it if it exists? what if it has been modified?

patricioe: oh I see

lib/splitclient-rb/managers/split_manager.rb
```diff
@@ -22,8 +22,11 @@ def splits
 
       @splits_repository.splits.each_with_object([]) do |(name, split), memo|
         split_model = Engine::Models::Split.new(split)
+        split_view = build_split_view(name, split)
 
-        memo << build_split_view(name, split) unless split_model.archived?
+        next if split_view[:name] == nil
```
patricioe: why would the name be null?

maksimf: @patricioe That is a very good question. I've spent quite some time to figure that out, but since we need to release new version I just added this check. So, I don't really know why there is a split with `nil` name in the cache.

PR #111, opened 2017-01-07
Change number

patricioe:


lib/splitclient-rb/clients/split_client.rb
```diff
@@ -62,22 +63,22 @@ def get_treatment(key, split_name, attributes = nil, split_data = nil, store_imp
       end
 
       start = Time.now
-      treatment_with_label = { label: Engine::Models::Label::EXCEPTION, treatment: Treatments::CONTROL }
+      treatment_label_change_number = { label: Engine::Models::Label::EXCEPTION, treatment: Treatments::CONTROL }
```
adilaijaz: So the one thing I'd call out is this:

how are you making sure that the right change number is reported even if there was an exception? Of course, if the exception was thrown while fetching the split (so that the split is null) then there can't be any change number.

I don't follow the code that well to be able to say that case is covered.

patricioe: good point. I'm waiting on @maksimf  to help on that front.

patricioe:
@maksimf this has passed now my manual tests. Please merge if you think it's good to go.

lib/splitclient-rb/clients/split_client.rb
maksimf: @adilaijaz everywhere we can we just log an exception and continue execution, i.e. if `changeNumber` will be fetched it should be reported back to the Split API

adilaijaz: @maksimf i don't think i follow

maksimf: @adilaijaz Let's divide possible exceptions into two sets:
1. Those exceptions which are not related to the successful delivery of impressions
2. Those which are related, i.e. if there was an exception while fetching split data, obviously, SDK can't report anything reasonable as impression

So, the answer to your initial question is: if we are talking about exceptions from the first set, then, even if an exception was thrown we will deliver an impression. But, we can't send impression if we're dealing with the exception of the second set.

If my answer still makes no sense, maybe I did not understand your question correctly.

PR #112, opened 2017-01-09
[WIP] Log exception if network error

maksimf:


maksimf:
@patricioe That was WIP, but I'll create one more to finish this one then

PR #114, opened 2017-01-13
Do not send bucketing key if nil

maksimf:


maksimf:
@patricioe That's quite an important change

maksimf:
@patricioe So, that's the main reason why half of the QOS tests were failing. As Fernando explained to me: at some point he changed the tests in a way that now they checked if they do not pass `bucketing_key`, then they expect it to be `null` in impression. What I did before was just copying it from the `matching_key` if it was `nil`.

PR #116, opened 2017-01-16
Use SCAN instead of KEYS

maksimf:


patricioe:
@maksimf is there any other corner cases scenario you may want to cover with integration tests here?

maksimf:
@patricioe I can't think of anything right now, do you mean anything in particular?

PR #117, opened 2017-01-17
Remove sized queue adapter

maksimf:


maksimf:
@patricioe it requires Ruby >= 2.2, we should support 2.0 and 2.1 as well

PR #120, opened 2017-01-31
Fix redis namespace

maksimf:


lib/splitclient-rb/split_config.rb
```diff
@@ -29,7 +29,7 @@ def initialize(opts = {})
       @events_uri = (opts[:events_uri] || SplitConfig.default_events_uri).chomp('/')
       @mode = opts[:mode] || SplitConfig.default_mode
       @redis_url = opts[:redis_url] || SplitConfig.default_redis_url
-      @redis_namespace = opts[:redis_namespace] || SplitConfig.default_redis_namespace
+      @redis_namespace = opts[:redis_namespace] ? "#{opts[:redis_namespace]}/#{SplitConfig.default_redis_namespace}" : SplitConfig.default_redis_namespace
```
patricioe: are we adding a slash? 

@sarrubia

patricioe: My understanding according to the spec is: `userPrefix.SPLITIO.blablabla` and i don't see that here.

maksimf: @patricioe @sarrubia 
There is a little ambiguity in the [docs](https://docs.google.com/document/d/10NSU3Fgmfa74ElM2mZGTqdNjwPG5mHmd5XSJG8X3nkA/edit), both `/` and `.` is mentioned there. What should I use?

maksimf:
@patricioe We're testing name space as well now

lib/splitclient-rb/split_config.rb
sarrubia: @maksimf @patricioe we are using a mix between dot and slash. Let me show a sample pattern to clarify it.

`userPrefix.SPLITIO/{sdk-language-version}/{instance-id}/impressions.{featureName}`

As you can see the slash is used to separate the `sdk-language-version` and the` instance-id`

PR #122, opened 2017-02-09
Fix redis namespace issue

maksimf:


maksimf:
@patricioe Do you mean the issue when there was an extra prefix (`RUBY-VERSION/MACHINE-IP`) to it? If so, then yes

lib/cache/repositories/repository.rb
```diff
@@ -12,7 +12,7 @@ def string(key)
       protected
 
       def namespace_key(key)
-        "#{@config.redis_namespace}/#{@config.machine_ip}/#{key}"
+        "#{@config.redis_namespace}/#{key}"
```
patricioe: @maksimf please look at the spec. It's not slash, it's a dot.

Slash is only for impressions and metrics data.

According to docs, till, splits and segment data:
```
Current till field:
SPLITIO.split.till
Fixed till field:
SPLITIO.splits.till

Current segment content:
SPLITIO.segmentData.{segmentName}
Fixed segment content:
SPLITIO.segment.{segmentName}

Current ready flags content:
SPLITIO.cache.ready.splits
SPLITIO.cache.ready.segments
Fixed cache flags for consistency:
SPLITIO.splits.ready
SPLITIO.segments.ready

Current fields which doesn’t change:
SPLITIO.split.{splitName}
SPLITIO.segments.registered
SPLITIO.segment.{segmentName}.till
This fields keeps working in the same way as defined in the Spec.```

PR #124, opened 2017-02-15
Allow non-string keys in get_treatment

maksimf:


lib/splitclient-rb/clients/split_client.rb
```diff
@@ -107,9 +107,9 @@ def get_treatment(key, split_name, attributes = nil, split_data = nil, store_imp
     def keys_from_key(key)
       case key.class.to_s
       when 'Hash'
-        key.values_at(:bucketing_key, :matching_key).map(&:to_s)
-      when 'String'
-        [nil, key.to_s]
+        key.values_at(:bucketing_key, :matching_key).map { |k| k.nil? ? nil : k.to_s }
+      else
+        [nil, key].map { |k| k.nil? ? nil : k.to_s }
```
patricioe: will String: "1" and Int: 1 hash to the same hash value?

patricioe:
conflicts too

lib/splitclient-rb/clients/split_client.rb
maksimf: Yep, is this what's expected?

maksimf:
@patricioe Fixed conflicts

lib/splitclient-rb/version.rb
```diff
@@ -1,3 +1,3 @@
 module SplitIoClient
-  VERSION = '3.2.3'
+  VERSION = '3.2.4'
```
patricioe: @maksimf max please drop this change from here so it does not override of conflict with latest head.

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -107,9 +107,9 @@ def get_treatment(key, split_name, attributes = nil, split_data = nil, store_imp
     def keys_from_key(key)
       case key.class.to_s
       when 'Hash'
-        key.values_at(:bucketing_key, :matching_key)
-      when 'String'
-        [nil, key]
+        key.values_at(:bucketing_key, :matching_key).map { |k| k.nil? ? nil : k.to_s }
+      else
+        [nil, key].map { |k| k.nil? ? nil : k.to_s }
```
patricioe: so

[number].to_s and "string".to_s have the same output? just checking.

maksimf:
@patricioe Yes, same

maksimf:
@patricioe conflicts fixed

lib/splitclient-rb/version.rb
```diff
@@ -1,3 +1,3 @@
 module SplitIoClient
-  VERSION = '3.2.4-rc2'
+  VERSION = '3.2.5'
```
patricioe: please drop this. Let's change version only during releases.

maksimf:
@patricioe got it, done

patricioe:
merged thanks

PR #126, opened 2017-02-21
Make producer work with all consumers

maksimf:


lib/cache/repositories/impressions/redis_repository.rb
```diff
@@ -39,14 +39,15 @@ def add_bulk(key, bucketing_key, treatments_labels_change_numbers, time)
           # delete fetched impressions afterwards
           def clear
             impressions = impression_keys.each_with_object([]) do |key, memo|
-              _, _, ip, = key.split('/')
+              _, _, ip, _ = key.split('/')
```
patricioe: too much magic and not very readable.

Mind adding a comment indicating a sample of the input so it's easier for reviews to see what's going on?

maksimf: yep, of course, moreover it's incorrect here

maksimf: @patricioe is it ok if I change that for a more readable option?

```ruby
ip = key.split('/')[-2]
```

lib/cache/repositories/impressions/redis_repository.rb
```diff
@@ -39,7 +39,7 @@ def add_bulk(key, bucketing_key, treatments_labels_change_numbers, time)
           # delete fetched impressions afterwards
           def clear
             impressions = impression_keys.each_with_object([]) do |key, memo|
-              _, _, ip, _ = key.split('/')
+              ip = key.split('/')[-2]
```
patricioe: add still a comment of a sample string you are parsing for the reviewers and for the posterity.

maksimf: @patricioe done

lib/cache/repositories/impressions/redis_repository.rb
```diff
@@ -39,14 +39,15 @@ def add_bulk(key, bucketing_key, treatments_labels_change_numbers, time)
           # delete fetched impressions afterwards
           def clear
             impressions = impression_keys.each_with_object([]) do |key, memo|
-              _, _, ip, = key.split('/')
+              ip = key.split('/')[-2] # 'prefix/sdk_lang/ip/impressions.name' -> ip
```
patricioe: what happens if there is an exception parsing? can we catch, continue with next key, clear the cache anyway and perhaps log a single line per batch?

maksimf: Done

lib/cache/repositories/impressions/redis_repository.rb
```diff
@@ -39,14 +38,19 @@ def add_bulk(key, bucketing_key, treatments_labels_change_numbers, time)
...
+              if ip.nil?
+                @config.logger.warn("Impressions IP parse error for key: #{key}")
```
patricioe: will the key still be deleted?

lib/cache/repositories/impressions/redis_repository.rb
patricioe: what does -1 do here?

maksimf: `[-2]` asks for the second element, but it count from right to left
`a = [1, 2, 3]`
`a[-2] # => 2`

lib/cache/repositories/impressions/redis_repository.rb
maksimf: yes if `ip` would be nil I call `next`, i.e. next iteration

PR #128, opened 2017-03-01
Provide interface to run SplitAdapter

maksimf:


Detailed-README.md
```diff
@@ -372,6 +372,40 @@ Note: options passed through cli have higher priority than those specified in th
...
+
+Other servers should work fine as well, but haven't been tested.
+
+### Unicorn
+
+If you're using Unicorn in the `memory` mode you'll need to include this line in your unicorn config (probably `config/unicorn.rb`):
+
+```ruby
+after_fork do |server, worker|
+  Rails.configuration.split_factory.run_adapter! if worker.nr > 0
```
patricioe: can we call it `resume` instead?

it, for now, can be an alias to `start` or `initialize` but it gives us something to evolve it in the future.

Detailed-README.md
```diff
@@ -372,6 +372,40 @@ Note: options passed through cli have higher priority than those specified in th
...
+end
+```
+
+This piece of code will resurrect threads after Unicorn forked new worker. You do not need this line if you have only one worker.
```
patricioe: > if you have only one worker
can that be detected by us and not push this decision to the user?

lib/splitclient-rb/split_factory.rb
```diff
@@ -17,12 +17,16 @@ def initialize(api_key, config_hash = {})
...
-      @adapter = SplitAdapter.new(@api_key, @config, @splits_repository, @segments_repository, @impressions_repository, @metrics_repository, @sdk_blocker)
+      @adapter = run_adapter!
 
       @client = SplitClient.new(@api_key, @config, @adapter, @splits_repository, @segments_repository, @impressions_repository, @metrics_repository)
       @manager = SplitManager.new(@api_key, @config, @adapter, @splits_repository)
 
       @sdk_blocker.block if @config.block_until_ready > 0
     end
+
+    def run_adapter!
```
patricioe: see comment before.

I'd have two methods: `start` and `resume`.

maksimf:
@patricioe got it

Detailed-README.md
maksimf: We're doing that already, just thought it's worth noticing, I'll remove that line

maksimf:
@patricioe Done

lib/splitclient-rb/split_factory.rb
```diff
@@ -17,12 +17,18 @@ def initialize(api_key, config_hash = {})
       @metrics_repository = MetricsRepository.new(@config.metrics_adapter, @config)
 
       @sdk_blocker = SDKBlocker.new(@config, @splits_repository, @segments_repository)
-      @adapter = SplitAdapter.new(@api_key, @config, @splits_repository, @segments_repository, @impressions_repository, @metrics_repository, @sdk_blocker)
+      @adapter = start!
```
patricioe: just for my own education, what does `!` mean?

maksimf: @patricioe  In Ruby we add bang to the end of the method name when it is "dangerous". Often it means that it changes the state, e.g. we have 2 methods `map` and `map!`. First one will return a modified copy, whereas the second one will modify the object itself. By adding bang to this method I want to show that this method changes the state as well - it changes cache. This 100% not required, but considered good coding style

PR #131, opened 2017-03-06
Disable http cache

maksimf:


maksimf:
@patricioe @adilaijaz Do you think this is necessary?

patricioe:
This is not necessary I don't think. Closing the PR for now.

PR #132, opened 2017-03-07
Log impression errors

maksimf:


lib/cache/repositories/impressions/redis_repository.rb
```diff
@@ -59,13 +59,17 @@ def clear
...
+          rescue StandardError => e
+            @config.logger.error("Exception while clearing impressions cache: #{e}")
           end
 
           private
 
           # Get all sets by prefix
           def impression_keys
             @adapter.find_sets_by_prefix("#{@config.redis_namespace}/*/impressions.*")
+          rescue StandardError => e
```
patricioe: what happens to the caller of this method if you catch the error and continue as you are doing here?

patricioe: `impressions = impression_keys.each_with_object([]) do |key, memo|`

maksimf: @patricioe `impression_keys` will return `nil` and no iteration would happen, i.e. `clear` method will return `nil` as well

PR #133, opened 2017-03-13
Add version spec to concurrent-ruby gem

punnie:
Because of its use of Concurrent::Map, this gem requires concurrent-ruby to be at least on version 1.0.0.

Not declaring this causes awkward scenarios where that dependency is not satisfied and the gem fails at runtime when not finding the missing concurrent-ruby classes.

### DIFF
--- a/splitclient-rb.gemspec
+++ b/splitclient-rb.gemspec
@@ -30,7 +30,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency "json", "~> 1.8"
   spec.add_runtime_dependency "thread_safe"
-  spec.add_runtime_dependency "concurrent-ruby"
+  spec.add_runtime_dependency "concurrent-ruby", "~> 1.0.0"
   spec.add_runtime_dependency "faraday"
   spec.add_runtime_dependency "faraday-http-cache"
   spec.add_runtime_dependency "faraday_middleware"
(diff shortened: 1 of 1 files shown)


maksimf:
@punnie What scenario exactly? Could you clarify a bit more, please?

punnie:
Included this gem in a project that had an existing dependency on concurrent-ruby `~> 0.8.0`. Since this gem doesn't declare that it needs a higher version only upon deployment—in staging, fortunately—and in runtime I found out the `Concurrent::Map` class was missing.

That class is only available from `1.0.0pre` on concurrent-ruby anyway, so if this gem uses it then it should declare a minimum version.

maksimf:
@punnie Makes sense, but why `~> 1.0.0` then? Maybe `>= 1.0.0` then?

punnie:
Up to you. 😉 

But be mindful that if the `>=` switch is used, any version above `1.0.0` will satisfy the dependency—as opposed with the `~>` switch where only versions that respect `1.0.X` will be accepted as a valid dependency. 

So if concurrent-ruby `2.0` or `1.1` gets releases and brings breaking changes there may be breakage when using the `>=` switch. If the `~>` switch is used then they won't be valid candidates.

maksimf:
@punnie Could you update the PR to be `~> 1.0`? I think we'll be fine with `1.x` versions

maksimf:
Thanks!

PR #135, opened 2017-03-17
Negation matcher fix

maksimf:


CHANGES.txt
```diff
@@ -1,3 +1,6 @@
+3.2.6
```
patricioe: Please rebase. We are still a 3.2.4-rc3

lib/splitclient-rb/version.rb
```diff
@@ -1,3 +1,3 @@
 module SplitIoClient
-  VERSION = '3.2.4-rc2'
+  VERSION = '3.2.6'
```
patricioe: discard this change please.

lib/engine/parser/condition.rb
```diff
@@ -38,20 +36,24 @@ def matchers
       @data[:matcherGroup][:matchers]
     end
 
-    def matcher_all_keys params
+    def negation_matcher(matcher)
+      NegationMatcher.new(matcher)
+    end
+
+    def matcher_all_keys(_params)
       AllKeysMatcher.new
```
patricioe: can we reuse the object?

maksimf: @patricioe for sure, done!

PR #137, opened 2017-03-18
improve logging

patricioe:


maksimf:
@patricioe Could you please add [this file](https://github.com/splitio/ruby-client/blob/c2ead9a72c2d451189b949d41eb73f39b6e6cb6f/build) to your global `gitignore` file or `bash_profile` as an alias? I believe it shouldn't be in the repo.

patricioe:
hey @maksimf , that was meant to be in the repo. It's a shortcut to build the gem.

maksimf:
@patricioe I got it, but we usually don't put this kind of stuff in the gem's repo

maksimf:
If you 100% need this in repo, then let's keep it, but if possible I'd rather not have this in the repo

PR #138, opened 2017-03-21
Fix impressions keys

maksimf:


lib/splitclient-rb/version.rb
```diff
@@ -1,3 +1,3 @@
 module SplitIoClient
-  VERSION = '3.2.4-rc3'
+  VERSION = '3.2.4-rc4'
```
patricioe: please drop this version change. I bump it up during the release.

PR #140, opened 2017-03-21
Fix impressions in memory issue

maksimf:


patricioe:
Merging now, but please consider my comment.

maksimf:
@patricioe I will definitely need to add some tests

PR #142, opened 2017-03-23
Allow configurable impressions/metrics slice size

maksimf:


lib/cache/repositories/impressions/memory_repository.rb
```diff
@@ -8,27 +8,28 @@ def initialize(adapter, config)
             @config = config
           end
 
-          # Store impression data in the selected adapter
+          # Store impression data in the memory adapter if choice
```
patricioe: > if choice
typo there?

lib/cache/repositories/impressions/redis_repository.rb
```diff
@@ -18,16 +16,17 @@ def add(split_name, data)
             )
           end
 
-          def add_bulk(key, bucketing_key, treatments_labels_change_numbers, time)
+          # Store impressions in bulk
+          def add_bulk(key, bucketing_key, treatments, time)
             @adapter.redis.pipelined do
-              treatments_labels_change_numbers.each_slice(IMPRESSIONS_SLICE) do |treatments_labels_change_numbers_slice|
-                treatments_labels_change_numbers_slice.each do |split_name, treatment_label_change_number|
+              treatments.each_slice(@config.impressions_slice_size) do |treatments_slice|
```
patricioe: this is when we store in redis/cache right?

how about the slice when we read from redis and post to Split?

lib/cache/repositories/impressions/memory_repository.rb
maksimf: Fixed

lib/cache/repositories/impressions/redis_repository.rb
maksimf: @patricioe we have `@config.impressions_queue_size` for that:
https://github.com/splitio/ruby-client/blob/development/lib/cache/repositories/impressions/redis_repository.rb#L47

patricioe: then in which case will we have more than 1000 treatments to store? I'm confused.

maksimf: @patricioe I didn't get the question

SenhorCastor:
Closing this, addressed in other PR (207)

PR #145, opened 2017-04-13
Set matchers

maksimf:


lib/engine/matchers/part_of_set_matcher.rb
```diff
@@ -0,0 +1,15 @@
...
+    def self.matcher_type
+      'PART_OF_SET'.freeze
+    end
+
+    def initialize(remote_array)
+      @remote_set = remote_array.to_set
+    end
+
+    def match?(_key, local_array)
+      local_array.to_set.intersect? @remote_set
```
patricioe: if the local_array is what the dev passed along in the code, make sure that:

`local_array intersect @remote` is not `true` when `local_array` is empty.

spec/engine/matchers/part_of_set_matcher_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  let(:split_json) { File.read(File.expand_path(File.join(File.dirname(__FILE__), '../../test_data/splits/part_of_set_matcher/split.json'))) }
+
+  before do
+    stub_request(:get, 'https://sdk.split.io/api/splitChanges?since=-1')
+      .to_return(status: 200, body: split_json)
+  end
+
+  it 'validates the treatment is ON for correct attribute value' do
+    expect(subject.get_treatment('user', 'LOCAL_IS_PART_OF', %w(a b c))).to eq SplitIoClient::Treatments::ON
```
patricioe: Do other matchers test are done this way with JSON? doesn't it become more testeable if you test the matcher in isolation? like array (input, operation, output) =>  asesertion?

create a test with empty array, the matcher should be false.

and one with null/nil just to see if you get control

maksimf: > Do other matchers test are done this way with JSON? doesn't it become more testeable if you test the matcher in isolation? like array (input, operation, output) => asesertion?
Makes sense, I've changed that to test matchers in isolation

maksimf:
@patricioe Ready for review

lib/engine/matchers/contains_any_matcher.rb
```diff
@@ -0,0 +1,15 @@
...
+    def self.matcher_type
+      'CONTAINS_ANY'.freeze
+    end
+
+    def initialize(attribute, remote_array)
+      super(attribute, remote_array)
+    end
+
+    def match?(_key, data)
+      local_set(data, @attribute).intersect? @remote_set
```
patricioe: what happens is local set is empty?
@SenhorCastor should we expect false?

patricioe: do we need a ` return false if @remote_set.empty?` ?

maksimf: No, here's a test for that: https://github.com/splitio/ruby-client/blob/feature/set_matchers/spec/engine/matchers/contains_any_matcher_spec.rb#L13

maksimf:
@patricioe I've updated this PR with new string matchers, ready to be reviewed/merged

PR #150, opened 2017-05-12
Impression listener

maksimf:
- [ ] add more specs

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,6 @@
+4.3.0 (August 7, 2017)
+- Add impression listener
+
 4.2.3 (August 4, 2017)
 - Use ENV vars in producer
 
--- a/lib/cache/routers/impression_router.rb
+++ b/lib/cache/routers/impression_router.rb
@@ -44,11 +44,5 @@ module SplitIoClient
         end
       end
     end
-
-    def random_interval(interval)
-      random_factor = Random.new.rand(50..100) / 100.0
-
-      interval * random_factor
-    end
   end
 end
--- a/lib/cache/routers/impression_router.rb
+++ b/lib/cache/routers/impression_router.rb
@@ -1,0 +1,54 @@
+module SplitIoClient
+  class ImpressionRouter
+    attr_reader :router_thread
+
+    def initialize(config)
+      @config = config
+      @listener = config.impression_listener
+      @queue = Queue.new
+      router_thread
+
+      if defined?(PhusionPassenger)
+        PhusionPassenger.on_event(:starting_worker_process) do |forked|
+          router_thread if forked
+        end
+      end
+    end
+
+    def add(impression)
+      @queue.push(impression)
+    end
+
+    def add_bulk(impressions)
+      impressions[:split_names].each do |split_name|
+        @queue.push(
+          split_name: split_name.to_s,
+          matching_key: impressions[:matching_key],
+          bucketing_key: impressions[:bucketing_key],
+          treatment: {
+            label: impressions[:treatments_labels_change_numbers][split_name.to_sym][:label],
+            treatment: impressions[:treatments_labels_change_numbers][split_name.to_sym][:treatment],
+            change_number: impressions[:treatments_labels_change_numbers][split_name.to_sym][:change_number]
+          },
+          attributes: impressions[:attributes]
+        )
+      end
+    end
+
+    private
+
+    def router_thread
+      Thread.new do
+        loop do
+          @listener.log(@queue.pop)
+        end
+      end
+    end
+
+    def random_interval(interval)
+      random_factor = Random.new.rand(50..100) / 100.0
+
+      interval * random_factor
+    end
+  end
+end
--- a/lib/splitclient-rb/cache/routers/impression_router.rb
+++ b/lib/splitclient-rb/cache/routers/impression_router.rb
@@ -40,7 +40,11 @@ module SplitIoClient
     def router_thread
       Thread.new do
         loop do
-          @listener.log(@queue.pop)
+          begin
+            @listener.log(@queue.pop)
+          rescue StandardError => error
+            @config.log_found_exception(__method__.to_s, error)
+          end
         end
       end
     end
--- a/lib/splitclient-rb/clients/split_client.rb
+++ b/lib/splitclient-rb/clients/split_client.rb
@@ -89,13 +89,8 @@ module SplitIoClient
 
         store_impression(
           split_name, matching_key, bucketing_key,
-<<<<<<< HEAD
-          { treatment: SplitIoClient::Engine::Models::Treatment::CONTROL, label: Engine::Models::Label::EXCEPTION },
-          store_impressions
-=======
-          { treatment: SplitIoClient::Treatments::CONTROL, label: Engine::Models::Label::EXCEPTION },
+          { treatment: SplitIoClient::Engine::Models::Treatment::CONTROL, label: SplitIoClient::Engine::Models::Label::EXCEPTION },
           store_impressions, attributes
->>>>>>> Add ImpressionListener support
         )
 
         return parsed_treatment(multiple, treatment_data)
@@ -104,11 +99,7 @@ module SplitIoClient
       begin
         latency = (Time.now - start) * 1000.0
         # Disable impressions if @config.impressions_queue_size == -1
-<<<<<<< HEAD
-        split && store_impression(split_name, matching_key, bucketing_key, treatment_data, store_impressions)
-=======
-        split && store_impression(split_name, matching_key, bucketing_key, treatment_label_change_number, store_impressions, attributes)
->>>>>>> Add ImpressionListener support
+        split && store_impression(split_name, matching_key, bucketing_key, treatment_data, store_impressions, attributes)
 
         # Measure
         @adapter.metrics.time('sdk.get_treatment', latency)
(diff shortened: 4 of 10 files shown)


maksimf:
@patricioe This PR is ready for review

lib/cache/routers/impression_router.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def router_thread
+      Thread.new do
+        loop do
+          @listener.log(@queue.pop)
+        end
+      end
+    end
+
+    def random_interval(interval)
```
patricioe: where is this used?

lib/cache/routers/impression_router.rb
```diff
@@ -0,0 +1,54 @@
...
+        )
+      end
+    end
+
+    private
+
+    def router_thread
+      Thread.new do
+        loop do
+          @listener.log(@queue.pop)
```
patricioe: what happens is there are exceptions here?

patricioe: also, does it block if there are no impressions?

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -120,6 +123,30 @@ def store_impression(split_name, matching_key, bucketing_key, treatment, store_i
       )
     end
 
+    def route_impression(split_name, matching_key, bucketing_key, treatment, attributes)
```
patricioe: no labels here?

lib/splitclient-rb/split_config.rb
```diff
@@ -251,6 +257,10 @@ def self.default_impressions_refresh_rate
       60
     end
 
+    def self.default_impression_listener_refresh_rate
```
patricioe: Please make the default to be immediate

lib/cache/routers/impression_router.rb
maksimf: > what happens is there are exceptions here?
thread dies, it doesn't affect the main thread anyhow
> also, does it block if there are no impressions?
when they are no impressions thread just sleeps

lib/cache/routers/impression_router.rb
maksimf: Nowhere, deleted

lib/splitclient-rb/clients/split_client.rb
maksimf: label is stored inside `treatment[:label]`

lib/splitclient-rb/split_config.rb
maksimf: Done

lib/cache/routers/impression_router.rb
patricioe: let's not let the thread dies. What happens after it dies? does it get created again?

maksimf: Nope, it just dies and that's it.

Do you want me to do the same thing we do with other threads? i.e. log all exceptions in the infinite loop?

patricioe: Yeah. If the thread dies it represents a problem for us.

Log should be warn.

CHANGES.txt
```diff
@@ -1,3 +1,6 @@
+4.3.0 (July 12, 2017)
+- Add impression listener
+
```
patricioe: Let's rebase this branch with latest develpment

lib/splitclient-rb/cache/routers/impression_router.rb
```diff
@@ -0,0 +1,48 @@
...
+        )
+      end
+    end
+
+    private
+
+    def router_thread
+      Thread.new do
+        loop do
+          @listener.log(@queue.pop)
```
patricioe: what happens if there is an exception here? will the thread die?

lib/splitclient-rb/cache/routers/impression_router.rb
```diff
@@ -0,0 +1,48 @@
...
+            change_number: impressions[:treatments_labels_change_numbers][split_name.to_sym][:change_number]
+          },
+          attributes: impressions[:attributes]
+        )
+      end
+    end
+
+    private
+
+    def router_thread
```
patricioe: what happens if there are no listeners? Can that happen?

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -116,7 +112,9 @@ def get_treatment(
       parsed_treatment(multiple, treatment_data)
     end
 
-    def store_impression(split_name, matching_key, bucketing_key, treatment, store_impressions)
+    def store_impression(split_name, matching_key, bucketing_key, treatment, store_impressions, attributes)
+      route_impression(split_name, matching_key, bucketing_key, treatment, attributes) if @config.impression_listener && store_impressions
+
       return if @config.impressions_queue_size <= 0 || !store_impressions
 
       @impressions_repository.add(split_name,
```
patricioe: impressions are stored even if the listener is not set , correct?

spec/cache/routers/impression_router_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'spec_helper'
+
+describe SplitIoClient::ImpressionRouter do
+  let(:dbl) { double }
```
patricioe: what is double?

maksimf: It's a dummy object which accepts any method on it, useful in tests

lib/splitclient-rb/cache/routers/impression_router.rb
maksimf: If there are no listeners, then we don't get here, see: https://github.com/splitio/ruby-client/blob/featute/impression_listener/lib/splitclient-rb/clients/split_client.rb#L116

`if @config.impression_listener && store_impressions`

lib/splitclient-rb/clients/split_client.rb
maksimf: yes, that's right

lib/splitclient-rb/cache/routers/impression_router.rb
```diff
@@ -0,0 +1,48 @@
...
+    attr_reader :router_thread
+
+    def initialize(config)
+      @config = config
+      @listener = config.impression_listener
+      @queue = Queue.new
+      router_thread
+
+      if defined?(PhusionPassenger)
+        PhusionPassenger.on_event(:starting_worker_process) do |forked|
```
patricioe: @maksimf do we need the same for unicorn and puma?

lib/splitclient-rb/cache/routers/impression_router.rb
patricioe: @maksimf did you address this issue? impressions listener thread should not die. Please use the same pattern as we use in the fetchers and regular impression sender.

lib/splitclient-rb/cache/routers/impression_router.rb
maksimf: No, for unicorn and puma we apply the fix from the Rails app side, there's also a ticket to switch to this approach for passenger: https://splitio.atlassian.net/browse/SPLIT-2604

lib/cache/routers/impression_router.rb
maksimf: Done

maksimf:
@patricioe Done

lib/splitclient-rb/cache/routers/impression_router.rb
patricioe: you sure? the link seems outdated.

I see `route_impression` being called on every Store_impressions. Mind helping understand the flow?

```ruby
    def store_impression(split_name, matching_key, bucketing_key, treatment, store_impressions, attributes)
      route_impression(split_name, matching_key, bucketing_key, treatment, attributes) if @config.impression_listener && store_impressions

      return if @config.impressions_queue_size <= 0 || !store_impressions

      @impressions_repository.add(split_name,
        'keyName' => matching_key,
        'bucketingKey' => bucketing_key,
        'treatment' => treatment[:treatment],
        'label' => @config.labels_enabled ? treatment[:label] : nil,
        'time' => (Time.now.to_f * 1000.0).to_i,
        'changeNumber' => treatment[:change_number]
      )
    end
```

patricioe: > what happens if there are no custom listeners?

maksimf: See the end of the line, I pointed out in the previous comment:
`if @config.impression_listener && store_impressions`
If `@config.impression_listener` is `nil`, i.e impression_listener is not set `route_impression` is not called

maksimf: In ruby you can wrirte something like this:
```ruby
do_this if that
```
Which is equal to:
```ruby
if that
  do_this
end
```

maksimf: So, the part of the code you mentioned is equal to:
```ruby
    def store_impression(split_name, matching_key, bucketing_key, treatment, store_impressions, attributes)
      if @config.impression_listener && store_impressions
        route_impression(split_name, matching_key, bucketing_key, treatment, attributes) 
      end

      return if @config.impressions_queue_size <= 0 || !store_impressions

      @impressions_repository.add(split_name,
        'keyName' => matching_key,
        'bucketingKey' => bucketing_key,
        'treatment' => treatment[:treatment],
        'label' => @config.labels_enabled ? treatment[:label] : nil,
        'time' => (Time.now.to_f * 1000.0).to_i,
        'changeNumber' => treatment[:change_number]
      )
    end
```

patricioe: my bad. I didn't see the end of the line.

Issue #151, opened 2017-05-19
Silence the logger when running rails console or when running tests


kamal:
By default, the SDK logs out to STDOUT. This can be very distracting and noisy when a developer starts a rails console session or when running tests. The SDK should detect these run modes and configure the default logger to be silent, or redirect to the Rails logger instead of STDOUT.

Here's a snippet of what I do in my split configuration.

```
options = {}
options[:logger] = NullObject.new if defined?(Rails::Console) || Rails.env.test?
...
SplitIoClient::SplitFactory.new(api_key, options)
```

Note: It's unfortunately that I have to explicitly pass in a `NullObject` which is not standard in Ruby. Perhaps the options parser should distinguish between passing in an explicit `nil` vs not configuring the value.

patricioe:
@maksimf let's discuss.

Thanks @kamal

maksimf:
Thank you for this, I think the suggestion makes sense and we can definitely go this way

maksimf:
@kamal done!

PR #152, opened 2017-05-19
Use Rails logger if available

maksimf:
#151


patricioe:
don't forget an entry in the changes file.

maksimf:
@patricioe done

PR #153, opened 2017-06-02
Dependency matcher

maksimf:


maksimf:
@patricioe Ready for review

lib/engine/partitions/treatments.rb
```diff
@@ -1,36 +0,0 @@
...
-    CONTROL = 'control'
-    OFF = 'off'
-    ON = 'on'
-
-    # get the actual value for the given treatment type
-    #
-    # @param type [string] treatment type
-    #
-    # @return [Treatment] treatment type value
-    def self.get_type(type)
```
maksimf: Yeah, I know. It's just used in tests (legacy), I can totally get rid of that

lib/engine/matchers/all_keys_matcher.rb
```diff
@@ -15,7 +15,7 @@ def initialize
     # @param key [string] key value to be matched
     #
     # @return [boolean] true for all instances
-    def match?(_key, _attributes)
+    def match?(_key, _split_treatment, _attributes)
```
maksimf: I see no reason why pass client here, client stores impressions/metrics and return evaluations, whereas treatment just return evaluations and we need only this

jwaimann: I think that if this is a different abstraction, its ok to pass an object that evaluates and just returns treatments. Maybe I would rename the parameter, to avoid confusion. split_treatment sounds more like the treatment value, instead of something that evaluates a split and returns a treatment

maksimf: @jwaimann I agree, I'm thinking of `SplitMatcher`, how does that sound to you?

jwaimann: Mmm I think that originally, we had a class Engine that does this work. Is it the engine module? Maybe that could be the name

jwaimann:
@maksimf : Question, where do you implement the logic for AttributeMatcher that appears on the spec?

maksimf:
@jwaimann I guess you're looking for this class: https://github.com/splitio/ruby-client/blob/development/lib/engine/parser/split_treatment.rb

lib/engine/matchers/all_keys_matcher.rb
NicoZelaya: Small comment, but I think the correct signature for the match function would be (key, evaluator, attributes).

spec/cache/repositories/impressions_repository_spec.rb
```diff
@@ -11,7 +11,9 @@
     let(:split_adapter) do
       SplitIoClient::SplitAdapter.new(nil, SplitIoClient::SplitConfig.new(mode: :nil), nil, nil, nil, nil, nil)
     end
-    let(:ip) { Socket.ip_address_list.detect { |intf| intf.ipv4_private? }.ip_address }
+    let(:ip) do
+      Socket.ip_address_list.detect { |intf| intf.ipv4_private? || intf.ipv4_loopback? }.ip_address
```
dendril: This looks as a code duplication. Could you please verify @maksimf?

spec/cache/senders/impressions_sender_spec.rb
```diff
@@ -7,7 +7,9 @@
     let(:repository) { SplitIoClient::Cache::Repositories::ImpressionsRepository.new(adapter, config) }
     let(:sender) { described_class.new(repository, config, nil) }
     let(:formatted_impressions) { sender.send(:formatted_impressions, repository.clear) }
-    let(:ip) { Socket.ip_address_list.detect { |intf| intf.ipv4_private? }.ip_address }
+    let(:ip) do
+      Socket.ip_address_list.detect { |intf| intf.ipv4_private? || intf.ipv4_loopback? }.ip_address
```
dendril: This looks as a code duplication. Could you please verify @maksimf?

spec/engine/matchers/depencdency_matcher_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(described_class.new('foo', %w(on yes true)).match?('foo', split_treatment, nil)).to eq(true)
+  end
+
+  it 'does not match' do
+    allow(split_treatment).to receive(:call).with({ matching_key: 'foo' }, 'foo', nil).and_return(treatment: 'no')
+
+    expect(described_class.new('foo', %w(on yes true)).match?('foo', split_treatment, nil)).to eq(false)
+  end
+
+  xit 'caches values' do
```
dendril: Is this test disabled?

spec/engine/parser/split_treatment_spec.rb
```diff
@@ -6,16 +6,33 @@
...
+  end
+
+  context 'dependency matcher' do
+    it 'uses cache' do
+      allow(split_treatment.instance_variable_get(:@splits_repository)).to receive(:get_split).and_return(split_data[:splits][0])
+
+      expect(split_treatment).to receive(:match).exactly(2).times
+      split_treatment.call({ bucketing_key: nil, matching_key: 'fake_user_id_1' }, split_data[:splits][0])
+
+      split_treatment.call({ bucketing_key: nil, matching_key: 'fake_user_id_1' }, split_data[:splits][1])
```
dendril: @maksimf could you please explain shortly what we are testing here?

lib/engine/matchers/all_keys_matcher.rb
maksimf: This is an idiom in Ruby: if I start param name with underscore it means that it's not used in the method (although required by the interface). So, since these params are not used in the method they require no docs here. Will add docs in `combining_matcher.rb` though

lib/engine/parser/split_treatment.rb
```diff
@@ -2,20 +2,38 @@ module SplitIoClient
   module Engine
     module Parser
       class SplitTreatment
-        def initialize(segments_repository)
+        def initialize(segments_repository, splits_repository, multiple = false)
+          @splits_repository = splits_repository
           @segments_repository = segments_repository
+          @multiple = multiple
+          @cache = {}
```
maksimf: Since new `@cache` instance gets created at each instantiation of `SplitTreatment`, I pass one `SplitTreatment` instance when `get_treatments` is called, hence cache is invalidated for each `get_treatments` / `get_treatment` call

lib/splitclient-rb/split_config.rb
```diff
@@ -334,9 +334,7 @@ def self.get_hostname
     #
     # @return [string]
     def self.get_ip
-      Socket.ip_address_list.detect { |intf| intf.ipv4_private? }.ip_address
-    rescue StandardError
-      'unknown'
+      Socket.ip_address_list.detect { |intf| intf.ipv4_private? || intf.ipv4_loopback? }.ip_address
```
maksimf: Nah, it's just a quick fix to make tests pass without internet access

spec/engine/matchers/depencdency_matcher_spec.rb
maksimf: yep

maksimf: Deleted it. I was thinking of putting cache test there, but ended up with `split_treatment_spec.rb`

spec/engine/parser/split_treatment_spec.rb
maksimf: We're testing that cache is actually being used.

This is how I do it: I have [2 splits](https://raw.githubusercontent.com/splitio/ruby-client/feature/dependency_matcher/spec/test_data/splits/engine/dependency_matcher.json): 
`test_whitelist` and `test_dependency`. I call `split_treatment` 2 times with first and second splits. 

If I had no cache I'd expect `match` method (which actually does the evaluation) to be called 3 times: first time for `test_whitelist`, second time for `test_dependency` and third time for `test_whitelist` again (since `test_dependency` depends on it). But, due to cache `match` is being called only 2 times - which exactly I would expect.

lib/engine/parser/split_treatment.rb
```diff
@@ -2,20 +2,38 @@ module SplitIoClient
...
         end
 
         def call(keys, split, attributes = nil)
+          # DependencyMatcher here, split is actually a split_name in this case
+          cache_result = split.is_a? String
+          split = @splits_repository.get_split(split) if cache_result
+          digest = Digest::MD5.hexdigest("#{{matching_key: keys[:matching_key]}}#{split}#{attributes}")
```
maksimf: Ok, then if you'll have results showing not so much people are actually make use of this cache I'll get rid of it, no problem

lib/engine/matchers/all_keys_matcher.rb
maksimf: I agree, evaluator suits best, done!

lib/engine/parser/split_treatment.rb
dendril: got it

lib/engine/parser/split_treatment.rb
dendril: Maybe you could work on that as well if you have time. cc @patricioe

PR #154, opened 2017-06-13
Client destroy

maksimf:
- [x] add logs when that method gets called
- [x] test responses of calling get_treatment and manager methods after the call to shutdown happened

### DIFF
--- a/lib/splitclient-rb/cache/adapters/memory_adapters/queue_adapter.rb
+++ b/lib/splitclient-rb/cache/adapters/memory_adapters/queue_adapter.rb
@@ -10,6 +10,11 @@ module SplitIoClient
             @current_size = Concurrent::AtomicFixnum.new(0)
           end
 
+          def clear(_ = nil)
+            @queue = Queue.new
+            @current_size = Concurrent::AtomicFixnum.new(0)
+          end
+
           # Adds data to queue in non-blocking mode
           def add_to_queue(data)
             fail ThreadError if @current_size.value >= @max_size
--- a/lib/splitclient-rb/cache/adapters/memory_adapters/queue_adapter.rb
+++ b/lib/splitclient-rb/cache/adapters/memory_adapters/queue_adapter.rb
@@ -12,7 +12,7 @@ module SplitIoClient
 
           def clear(_ = nil)
             @queue = Queue.new
-            @current_size = Concurrent::AtomicFixnum.new(0)
+            @current_size.value = 0
           end
 
           # Adds data to queue in non-blocking mode
--- a/lib/splitclient-rb/cache/senders/impressions_sender.rb
+++ b/lib/splitclient-rb/cache/senders/impressions_sender.rb
@@ -31,7 +31,7 @@ module SplitIoClient
         private
 
         def impressions_thread
-          Thread.new do
+          @config.threads[:impressions_sender] = Thread.new do
             @config.logger.info('Starting impressions service')
 
             loop do
--- a/lib/splitclient-rb/cache/senders/impressions_sender.rb
+++ b/lib/splitclient-rb/cache/senders/impressions_sender.rb
@@ -32,12 +32,18 @@ module SplitIoClient
 
         def impressions_thread
           @config.threads[:impressions_sender] = Thread.new do
-            @config.logger.info('Starting impressions service')
+            begin
+              @config.logger.info('Starting impressions service')
 
-            loop do
+              loop do
+                post_impressions
+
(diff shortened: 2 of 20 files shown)


maksimf:
Logs after `split_client.destroy` called:

```
5 splits retrieved. since=-1
storing split (ruby)
storing split (ruby_5)
storing split (ruby_4)
storing split (ruby_3)
storing split (ruby_1)
segments seen(0): []


Started GET "/destroy" for ::1 at 2017-06-15 18:45:44 +0300
  ActiveRecord::SchemaMigration Load (0.3ms)  SELECT "schema_migrations".* FROM "schema_migrations"
Processing by TestingRubyController#destroy as HTML
Shutting down split client...
Split client shutting down complete
```

maksimf:
@patricioe Ready to be merged

patricioe:
Are impressions being sent during the destroy process?

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -116,6 +116,19 @@ def get_treatment(
       parsed_treatment(multiple, treatment_data)
     end
 
+    def destroy
+      @config.logger.info('Split client shutdown started...') if @config.debug_enabled
+
+      @config.threads.each { |name, thread| Thread.kill(thread) }
```
patricioe: don't we need to send the impressions before shutdown?

maksimf:
@patricioe nope, all left impressions are dropped

lib/splitclient-rb/clients/split_client.rb
patricioe: @maksimf can we fire a last call before shutdown and make it NON blocking?

maksimf: @patricioe yes, will do

lib/splitclient-rb/cache/adapters/memory_adapters/queue_adapter.rb
```diff
@@ -10,6 +10,11 @@ def initialize(max_size)
             @current_size = Concurrent::AtomicFixnum.new(0)
           end
 
+          def clear(_ = nil)
+            @queue = Queue.new
+            @current_size = Concurrent::AtomicFixnum.new(0)
```
patricioe: why are you assigning a new variable? if the existing variable is atomic, why not setting it to zero?

lib/splitclient-rb/cache/senders/impressions_sender.rb
```diff
@@ -31,13 +31,19 @@ def call
...
-            @config.logger.info('Starting impressions service')
+          @config.threads[:impressions_sender] = Thread.new do
+            begin
+              @config.logger.info('Starting impressions service')
 
-            loop do
+              loop do
+                post_impressions
+
+                sleep(SplitIoClient::Utilities.randomize_interval(@config.impressions_refresh_rate))
```
patricioe: @maksimf remember that this cause an infinite loop before. If the post errors out, it skips the sleep and cycles again there is a nextwork issue. Please restablish the previous pattern.

lib/splitclient-rb/cache/senders/impressions_sender.rb
```diff
@@ -31,13 +31,19 @@ def call
...
+              end
+            rescue SplitIoClient::ImpressionShutdownException
               post_impressions
 
-              sleep(SplitIoClient::Utilities.randomize_interval(@config.impressions_refresh_rate))
+              @impressions_repository.clear
```
patricioe: You are accounting for the new impressions listener shutdown as well?

maksimf: Nope, are you ok if I'll fix that after Impression listener PR is merged?

lib/splitclient-rb/cache/senders/impressions_sender.rb
maksimf: @patricioe But we're catching exceptions inside `post_impressions` method:

```ruby
def post_impressions
  impressions_client.post
rescue StandardError => error
  @config.log_found_exception(__method__.to_s, error)
end
```

maksimf:
@patricioe Done

lib/splitclient-rb/cache/senders/impressions_sender.rb
patricioe: @maksimf I'm merging this. Mind addressing the impression listener shutdown right now? I'm merging this now.

patricioe:
merging this.

@maksimf you are missing the changelog entry.

PR #159, opened 2017-07-03
Update .gemspec to require faraday_middleware minor version

mvpgomes:
When building the :api_client in SplitIoClient::Api::Client,
the Faraday::Gzip is been used as the middleware for the Faraday
Client

```
def api_client
  @api_client ||= Faraday.new do |builder|
    builder.use FaradayMiddleware::Gzip
    builder.adapter :net_http_persistent
  end
end
```

However, if a project already have faraday_middleware installed
where the version is '< 0.10', http requests will fail,
since FaradayMiddleware::Gzip was introduced only in version '0.10'.
This commit updates .gemspec to require a minor version for
faraday_middleware '>= 0.10'.

Closes #159

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This SDK is designed to work with [Split](https://www.split.io), the platform fo
 
 ### Quick setup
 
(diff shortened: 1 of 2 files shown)


mvpgomes:
Thanks @maksimf 🎉

PR #161, opened 2017-07-03
Move files to be under lib/

### LINKED ISSUE #157: Gem Namespace & Directory Practice
Congrats, this gem looks really nice and I'm really looking forward to using this it with the Split.io service. 

While doing a technical review, I noticed that the gem has several namespace issues. These appear to be simple fixes but wanted your thoughts before doing a pull request. 

### General Directory & Namespace

Below are some reference articles on best practices on how to structure the gem lib directory. Having been bitten by these issues before, I consider these really important because file load conflicts are very hard to fix from the outside.

* https://blog.codeship.com/exploring-structure-ruby-gems/
* http://weblog.rubyonrails.org/2009/9/1/gem-packaging-best-practices/

The directories `lib/cache`, `lib/engine`, and `lib/exceptions` would need to be in `lib/splitcient-rb/...` to avoid conflicts. Good news is that each file has the proper module namespace, so a simple git mv for each of these directories followed by changes to paths in `sliptclient-rb.rb` should be sufficient.

### File Namespace

The other issue I see is that `lib/splitclient-rb_utilities.rb` defines a top level `Utilities` constant. This file name is OK since it is prefixed with the gem name, however, that module name is very ambiguous. Suggest changing the contents to use the `SplitIoClient::Utilities` namespace and making changes in the code to remove the `::` in places and everything should work fine.

What do you think? Is this something that you can fix or would you like a PR for it?
Again, thanks for the great work in the gem.

maksimf:
To avoid namespacing issues move all code inside lib/splitclient-rb/ directory

resolves #157


metaskills:
Thanks! I see that this work is in the `development` branch vs. `master`. I do not see a gem on rubygems.org either. Would you mind sharing how releases are done for this project?

maksimf:
@metaskills Sure, when we get the number of features merged and tested in `development` we merge development to master and tag it with new version, after that we update gem at rubygems.org. You can use `development` version if you want to be on the edge :-)

metaskills:
Thanks! I forgot I could check the name in the gemspec. Found the gem name on rubygems from there. https://rubygems.org/gems/splitclient-rb

Will bundle to development for our initial use.

PR #165, opened 2017-07-13
Remove extra dependencies

### LINKED ISSUE #158: HTTP requests failing when faraday_middleware version < 0.10
While configuring `splitclient-rb` for an existing application, I've bootstrap `SplitIoClient` following the available documention:

```ruby
require 'splitclient-rb'

options = {block_until_ready: 3}
split_factory = SplitIoClient::SplitFactoryBuilder.build("YOUR_API_KEY", options)
```
when instatiating the client I'm getting the following error:

```
ERROR -- : Failed to make a http request
INFO -- : Starting metrics service
ERROR -- : [splitclient-rb] Unexpected exception in store_splits: #<NoMethodError: undefined method `[]' for nil:NilClass> undefined method `[]' for nil:NilClass
	/bundle/gems/splitclient-rb-4.1.0/lib/cache/stores/split_store.rb:45:in `store_splits'
	/bundle/gems/splitclient-rb-4.1.0/lib/cache/stores/split_store.rb:35:in `block (2 levels) in splits_thread'
	/bundle/gems/splitclient-rb-4.1.0/lib/cache/stores/split_store.rb:34:in `loop'
	/bundle/gems/splitclient-rb-4.1.0/lib/cache/stores/split_store.rb:34:in `block in splits_thread'
INFO -- : Starting impressions service
WARN -- : uninitialized constant FaradayMiddleware::Gzip
```
If you look at the last line of the log, a warning is raised by `Faraday` saying that it doesn't know what is `FaradayMiddleware::Gzip`. **NOTE:** if you don't provide `options` to ` SplitIoClient::SplitFactoryBuilder.build`, when instantiating the client the request get stuck and we need to manually terminate the application execution.

After looking at `https://github.com/splitio/ruby-client/blob/master/lib/engine/api/client.rb#L10`, I saw that we are instantiating `Faraday` with `FaradayMiddleware::Gzip`, which is a dependency defined in `faraday_middleware` gem.

After digging my code I discovered that I already have installed `faraday_middleware` in version `0.9.2`. The issue is that  `FaradayMiddleware::Gzip` is only available from version `>= 0.10` https://github.com/lostisland/faraday_middleware/releases/tag/v0.10.0.

However, the currently `.gemspec` is not specifying a minor version for `faraday_middleware`, and if you already have a version installed, the gem will use that one, not matter which is the version of the installed gem. I will open a small PR to update `.gemspec` to specify a minor version for `faraday_middleware` as soon as possible 😄

maksimf:
resolves #158


maksimf:
@patricioe I've tested it with QOS (puma_mem), please review. If it's fine, then please squash all those commits

PR #172, opened 2017-09-04
Move passenger threads fix

maksimf:
I've assigned version 4.4.0 to this tiny fix to draw users' attention to it, this is potentially breaking changes. We'll need to notify all our customers, who use passenger add the fix manually, see: https://github.com/splitio/ruby-client/blob/feature/move_passenger_threads_fix/Detailed-README.md#phusion-passenger

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,7 @@
+4.4.0
+
+Extract passenger threads fix to the separate file
+
 4.3.0
 
 Add support for impression listener, there is an ability to use user-defined class to capture all impressions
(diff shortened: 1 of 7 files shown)


NEWS
```diff
@@ -1,3 +1,7 @@
+4.4.0
+
+Extract passenger threads fix to the separate file
```
patricioe: add what the users need to do to remain "compatible" with previous code. Pretty much, how to adjust their code.

maksimf: they will need to add this file: https://github.com/splitio/ruby-client/blob/feature/move_passenger_threads_fix/Detailed-README.md#phusion-passenger

PR #176, opened 2017-09-20
Matchers refactoring

maksimf:
## Recap
- All string matchers allow matching by matching_key when attribute is not present
- All matcher classes were refactored

## TODO

- [x] Add ANDing tests
- [x] Add string matchers tests, matching on both `matching_key` and `attribute`
- [x] Update version & changelog


spec/engine/matchers/attribute_matcher_spec.rb
```diff
@@ -1,55 +1,62 @@
...
-        expect(described_class.new('attr', matcher).match?('key', nil, nil, attr: 'value')).to eq(true)
-      end
-
-      it 'does not match' do
-        expect(described_class.new('attr', matcher).match?('key', nil, nil, attr: 'valuee')).to eq(false)
-      end
-    end
-  end
-end
+# describe SplitIoClient::AttributeMatcher do
```
patricioe: @maksimf what happened here?

PR #178, opened 2017-11-07
Cache hostname in Faraday

maksimf:
https://github.com/lostisland/faraday/issues/713#issuecomment-316462052

### DIFF
--- a/lib/splitclient-rb/engine/api/client.rb
+++ b/lib/splitclient-rb/engine/api/client.rb
@@ -6,7 +6,8 @@ module SplitIoClient
       RUBY_ENCODING = '1.9'.respond_to?(:force_encoding)
 
       def get_api(url, config, api_key, params = {})
-        api_client.get(url, params) do |req|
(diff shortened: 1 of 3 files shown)


lib/splitclient-rb/engine/api/client.rb
```diff
@@ -6,7 +6,8 @@ class Client
       RUBY_ENCODING = '1.9'.respond_to?(:force_encoding)
 
       def get_api(url, config, api_key, params = {})
-        api_client.get(url, params) do |req|
+        parsed_url = URI.parse(url)
```
patricioe: what is this for max?

maksimf: Long explanation is available from the link above. Short: I had an issue when I've seen phrase
`no_proxy is not supported` flooding logs (it appeared one time for 1 request to Split API), this is the solution.

SenhorCastor:
Closing this

Issue #180, opened 2017-11-07
Redis Availability


metaskills:
When we tested Split.io in `standalone` mode, I believe Redis going down caused `get_treatment` to not raise exceptions and return excluded. I could be wrong about this because as of now, we run the Go synchronizer and hence the Ruby apps are in `consumer` mode. Under this mode when the Redis connection goes down, any page that uses `get_treatment` will error out.

Our wrapper gem solves this by rescuing `Redis::BaseError` and returning `SplitIoClient::Engine::Models::Treatment::CONTROL`. To be fair, we never really had Redis go down and we do have our Split Redis AWS instance setup in such a way that we feel pretty confident about it being there for us. That said, just wanted to share this pattern and ask if there was anything more we could do on our end. Thanks in advance!

adilaijaz:
Thanks for flagging. That's not expected behavior. We are looking into this. Will keep you posted.

patricioe:
Ken,

This PR should address the issue: https://github.com/splitio/ruby-client/pull/184

Release coming soon but I wanted to let you know in case you need to test/use it immediately.

patricioe:
Ken,

The version 4.3.1 should fix the problem you ran into. I'll go ahead and close this ticket. Let us know if we can help with anything else.

metaskills:
Thanks! Confirmed this does indeed work.

PR #181, opened 2017-11-08
Fall gracefully when Redis is down

maksimf:
https://github.com/splitio/ruby-client/issues/180


patricioe:
Could you please a unit test that mocks the interaction with Redis and
throws an exception?

maksimf:
@patricioe done

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -148,6 +154,9 @@ def store_impression(split_name, matching_key, bucketing_key, treatment, store_i
         'time' => time,
         'changeNumber' => treatment[:change_number]
       )
+    rescue StandardError => error
+      @config.log_found_exception(__method__.to_s, error)
+      @config.logger.warn('Failed to store impression')
```
patricioe: did you forget to remove this?

you need to pass `Failed to store impression` to the method above or another way to provide a context message

PR #190, opened 2018-01-25
Add track()

maksimf:


patricioe:
remember to issue against development branch

maksimf:
fixed

patricioe:
@maksimf how are the events flush being trigger? I don't see that in your PR.

lib/splitclient-rb/cache/adapters/redis_adapter.rb
```diff
@@ -118,6 +118,23 @@ def random_set_elements(key, count)
           @redis.srandmember(key, count)
         end
 
+        # Queue
+        def add_to_queue(key, val, max_size)
+          @redis.del(key) if @redis.llen(key) >= max_size
```
sarrubia: @maksimf my concern here is that:  
 - if this delete key when max_size is reached is executed in redis-shared mode, the events into redis LIST will be removed and the consumer (split synchronizer) will lost those deleted events.

I think that eventsQueueSize configuration param is only to in-memory mode (not redis)

maksimf:
@patricioe this PR is `[WIP]`

lib/splitclient-rb/cache/adapters/redis_adapter.rb
maksimf: ok, thank you, I'll fix that

maksimf:
@patricioe please take a look

NEWS
```diff
@@ -1,3 +1,7 @@
+4.4.0
+
+Add track() API
```
patricioe: replace with "Add the ability to send event data to split service (.track() api)

lib/splitclient-rb/cache/repositories/events/memory_repository.rb
```diff
@@ -0,0 +1,28 @@
...
+
+          def initialize(adapter, config)
+            @adapter = adapter
+            @config = config
+          end
+
+          def add(key, traffic_type, event_type, time, value)
+            @adapter.add_to_queue(m: metadata, e: event(key, traffic_type, event_type, time, value))
+          rescue ThreadError # queue is full
+            @config.logger.warn("Dropping events. Current size is #{@config.events_queue_size}. " \
```
patricioe: can this be logged only if debug is enabled?

lib/splitclient-rb/cache/senders/events_sender.rb
```diff
@@ -0,0 +1,47 @@
...
+        end
+
+        def call
+          if ENV['SPLITCLIENT_ENV'] == 'test'
+            post_events
+          else
+            events_thread
+
+            if defined?(PhusionPassenger)
+              PhusionPassenger.on_event(:starting_worker_process) do |forked|
```
patricioe: what's the story with puma server here?

maksimf: Same thing as with every other thread we have: we're recreating threads for Passenger workers

lib/splitclient-rb/cache/adapters/redis_adapter.rb
```diff
@@ -118,6 +118,21 @@ def random_set_elements(key, count)
           @redis.srandmember(key, count)
         end
 
+        # Queue
+        def add_to_queue(key, val)
+          @redis.rpush(key, val)
+        end
+
+        def get_from_queue(key, count)
```
sarrubia: LGTM

maksimf:
@patricioe fixed

Issue #191, opened 2018-02-09
JRuby support


tjsousa:
I'm using the ruby client gem in one of our projects at Talkdesk and when running on JRuby an error is thrown due to a dependency using C extensions:

Just wondering if this is a known limitation and if there is value in being compatible with JRuby?

```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

current directory:
/usr/local/bundle/gems/digest-murmurhash-1.1.1/ext/digest/murmurhash
/opt/jruby/bin/jruby -r ./siteconf20180209-5-1h8ixdd.rb extconf.rb
creating Makefile

current directory:
/usr/local/bundle/gems/digest-murmurhash-1.1.1/ext/digest/murmurhash
make "DESTDIR=" clean

current directory:
/usr/local/bundle/gems/digest-murmurhash-1.1.1/ext/digest/murmurhash
make "DESTDIR="
make: *** No rule to make target '/opt/jruby/lib/ruby/include/ruby/ruby.h',
needed by 'neutral2.o'.  Stop.

make failed, exit code 2

Gem files will remain installed in
/usr/local/bundle/gems/digest-murmurhash-1.1.1 for inspection.
```

SenhorCastor:
Hi tjsousa, thanks for reaching out.

Currently we do not support JRuby - we used a library that is in C -. Our team is aware of this and we will work on making it compatible with JRuby in our next sprint.

I will keep this open and comment once we have an update

Thanks
Fernando

goldensun8891:
@tjsousa we have a RC branch that is now available to address this if you'd like to use it as a dry run?

patricioe:
@tjsousa ping.

Would it be possible for you to try this release candidate?

SenhorCastor:
Closing this, we have a JRuby gem available.

If you have any issues dont hesitate to contact

PR #193, opened 2018-02-23
Move C and Java implementation of murmurhash inside the gem

maksimf:
This PR is WIP

This solves #191


Rakefile
```diff
@@ -1,4 +1,24 @@
-require "bundler/gem_tasks"
-task :default => :spec
+require 'bundler/gem_tasks'
```
patricioe: @maksimf does this have any implications on our customers build tools?

Rakefile
```diff
@@ -1,4 +1,24 @@
...
+RSpec::Core::RakeTask.new(:spec)
 
-Dir.glob('tasks/**/*.rake').each(&method(:import))
+task :spec => :compile
+case RUBY_PLATFORM
+when 'java'
+  require 'rake/javaextensiontask'
+  Rake::JavaExtensionTask.new 'murmurhash' do |ext|
+    ext.lib_dir = 'lib/murmurhash'
+    ext.target_version = '1.7'
```
patricioe: is this = to 1.7 or can it be greater than 1.7?

what's the minimum version of java we can support?

lib/splitclient-rb/engine/evaluator/splitter.rb
```diff
@@ -50,7 +48,12 @@ def count_hash(key, seed, legacy)
       end
 
       def murmur_hash(key, seed)
-        Digest::MurmurHash3_x86_32.rawdigest(key, [seed].pack('L'))
+        case RUBY_PLATFORM
```
patricioe: what's the runtime penalty?

Can we avoid making this case statement on every evaluation and bind it one time during start up?

Rakefile
maksimf: What do you mean? I just changed default rake task not to be `spec`

Rakefile
maksimf: I had compilation issues with default version (1.6)

lib/splitclient-rb/engine/evaluator/splitter.rb
maksimf: Yes, thank you, I'll change that

maksimf: @patricioe fixed

Rakefile
patricioe: is 1.7 the minimum version? or does it have to match the exact version?
I want to make sure it will work with 1.8 as well.

maksimf: It is the minimum version. By default it was compiling for 1.5, but there were compilation issues and error for 1.5 and warnings for 1.6:

1.5:

```
javac  -target 1.5 -source 1.5 -Xlint:unchecked  -cp "/Users/max/.rbenv/versions/jruby-9.1.15.0/lib/jruby.jar:" -d tmp/java/murmurhash ext/murmurhash/MurmurHash3.java
warning: [options] bootstrap class path not set in conjunction with -source 1.5
error: Source option 1.5 is no longer supported. Use 1.6 or later.
error: Target option 1.5 is no longer supported. Use 1.6 or later.
```

1.6:

```
javac  -target 1.6 -source 1.6 -Xlint:unchecked  -cp "/Users/max/.rbenv/versions/jruby-9.1.15.0/lib/jruby.jar:" -d tmp/java/murmurhash ext/murmurhash/MurmurHash3.java
warning: [options] bootstrap class path not set in conjunction with -source 1.6
warning: [options] source value 1.6 is obsolete and will be removed in a future release
warning: [options] target value 1.6 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
```

So, I decided to use minimum of 1.7.

PR #198, opened 2018-04-26
Fix empty state issues when Redis is down

maksimf:


SenhorCastor:
Closing htis, it is tackled as part of other PR (currently 207)

PR #206, opened 2018-07-26
Turn Impression Listener into an optional SDK feature

ezesplit:
User feedback has made us aware that the documentation does not reflect whether using an impression listener is required as part of setting up the Split SDK. In fact, failure to do so, currently ends up in exceptions being raised in the `ImpressionRouter` class. 

This proposal turns impression listener into an optional feature, and prevents the impression thread from being started if a listener is not in place. Also, updates the `split_client` logic a bit, making it more readable, and delegating the check for an impression listener instance to the impression router class.

Finally, it updates the documentation to clarify that impression listener is now an optional feature.

### DIFF
--- a/Detailed-README.md
+++ b/Detailed-README.md
@@ -366,7 +366,9 @@ transport_debug_enabled: true # used for log transport data (mostly http request
 
 ### Impression Listener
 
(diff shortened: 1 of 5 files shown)


patricioe:
@SenhorCastor can we issue this against development?

LGTM

ezesplit:
@patricioe rebased to development.

PR #214, opened 2018-08-27
Fix incorrect API key causes unexpected exception

ezesplit:
When using an invalid API key the SDK raises a `NoMethodError: undefined method `[]' for nil:NilClass` exception that doesn't look too good. This PR addresses that by adding logic to treat this error and log a proper message indicating the possible cause of the issue.

Missing:
- Tests
- Fix Rubocop violations on new code


lib/splitclient-rb/engine/api/segments.rb
```diff
@@ -10,52 +15,53 @@ def initialize(api_key, config, metrics, segments_repository)
...
           since = @segments_repository.get_change_number(name)
-          while true
-            fetch_segments(name, prefix, since).each { |segment| @segments_repository.add_to_segment(segment) }
-            @config.logger.debug("Segment #{name} fetched before: #{since}, till: #{@segments_repository.get_change_number(name)}") if @config.debug_enabled
+          loop do
+            segment = fetch_segment_changes(name, since)
+            @segments_repository.add_to_segment(segment)
 
-            break if (since.to_i >= @segments_repository.get_change_number(name).to_i)
+            @config.logger.debug("Segment #{name} fetched before: #{since}, \
```
SenhorCastor: no if config.debug.enabled?

SenhorCastor: actually missing log_if_debug?

SenhorCastor:
Also move the version to the pre.rc that we talked

ignaciosplit:
@SenhorCastor feedback applied

PR #223, opened 2018-10-05
Cache Wrapper

ignaciosplit:
Cache wrapper when using Redis


CHANGES.txt
```diff
@@ -1,3 +1,6 @@
+5.1.3
+- Add cache wrapper to treatments and segments.
```
NicoZelaya:
```suggestion
- Add cache wrapper to splits and segments.
```

NEWS
```diff
@@ -1,3 +1,7 @@
+5.1.3
+
+Add cache wrapper to treatments and segments.
```
NicoZelaya:
```suggestion
Add cache wrapper to splits and segments. 
```

PR #227, opened 2018-11-22
Remove producer mode

ezesplit:
This PR  removes the producer mode and makes redis+consumer and memory+standalone the only valid SDK modes

It also fixes a small glitch caused by `api_post` wrapped having different return types based on internal logic


ezesplit:
Created in error. Wrong source and target branches

PR #228, opened 2018-11-22
Remove producer

ezesplit:
This PR removes the producer mode and makes redis+consumer and memory+standalone the only valid SDK modes

It also fixes a small glitch caused by api_post wrapped having different return types based on internal logic

[temporarily created to compare w/ impressions_revamp as source branch]

TODO: update documentation


lib/splitclient-rb/engine/api/events.rb
```diff
@@ -22,10 +22,12 @@ def post
             'SplitSDKVersion' => event_slice[0][:m][:s]
           )
 
-          if (200..299).include? result.status
-            SplitIoClient.configuration.logger.debug("Events reported: #{event_slice.size}") if SplitIoClient.configuration.debug_enabled
+          if response.success?
```
mredolatti: Does this .sucess? match any status code >=200 and < 300? Out events BE doesn't return 200 exactly, but sometimes 203, 204, etc

lib/splitclient-rb/engine/api/events.rb
```diff
@@ -22,10 +22,12 @@ def post
...
+            SplitLogger.log_if_debug("Events reported: #{event_slice.size}")
           else
-            SplitIoClient.configuration.logger.error("Unexpected status code while posting events: #{result.status}")
+            SplitLogger.log_error("Unexpected status code while posting events: #{response.status}." \
+            " - Check your API key and base URI")
+            raise 'Split SDK failed to connect to backend to post events'
```
mredolatti: Is there any reason for throwing here instead of just logging? Ideally we won't to avoid raising exceptions that may under any circumstance propagate to the user and bring down the application. My understanding is that this is within the synchronizer portion of the sdk. Is this guarded by a try/catch (or the ruby equivalent)? Unless there's a very specific reason to throw i'd rather just log the error as we are doing on every other sdk. cc: @NicoZelaya @mmelograno

mmelograno: agreed with you

lib/splitclient-rb/engine/api/impressions.rb
```diff
@@ -13,12 +13,14 @@ def post
         end
 
         impressions_by_ip.each do |ip, impressions|
-          result = post_api("#{SplitIoClient.configuration.events_uri}/testImpressions/bulk", @api_key, impressions, 'SplitSDKMachineIP' => ip)
+          response = post_api("#{SplitIoClient.configuration.events_uri}/testImpressions/bulk", @api_key, impressions, 'SplitSDKMachineIP' => ip)
 
-          if (200..299).include? result.status
-            SplitIoClient.configuration.logger.debug("Impressions reported: #{total_impressions(@impressions)}") if SplitIoClient.configuration.debug_enabled
+          if response.success?
```
mmelograno: consider on this line, the same as @mredolatti asked on top

lib/splitclient-rb/engine/api/events.rb
NicoZelaya: The check for the range is included on Faraday. Already discussed on site.

lib/splitclient-rb/engine/api/events.rb
NicoZelaya: They're catching that from a parent function.

PR #240, opened 2019-01-25
Passed in redis url is ignored and always defaults to REDIS_URL env var

BennySitbon:
The client allows to pass a `redis_url` in the init params, but never uses it.
 It saves the url in the `@redis_url` *instance* var of the split config. When the param is used, in the `SplitConfig. init_cache_adapter` it tries to reach to `@redis_url`, but since this method is in the class context, and not in the instance context, it will always be`nil`. Therefore passing `nil` to `SplitIoClient::Cache::Adapters::RedisAdapter.new`, which in turn eventually defaults to `REDIS_URL` config var in the `redis` client itself.

### DIFF
--- a/lib/splitclient-rb/split_config.rb
+++ b/lib/splitclient-rb/split_config.rb
@@ -43,7 +43,7 @@ module SplitIoClient
       @redis_url = opts[:redis_url] || SplitConfig.default_redis_url
       @redis_namespace = opts[:redis_namespace] ? "#{opts[:redis_namespace]}.#{SplitConfig.default_redis_namespace}" : SplitConfig.default_redis_namespace
       @cache_adapter = SplitConfig.init_cache_adapter(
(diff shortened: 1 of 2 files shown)


NicoZelaya:
Hi @BennySitbon ! 

First of all, thank you so much for flagging and taking the time to submit a PR! Our sincerest apologies for any inconveniences this may have caused. 
Our Ruby devs are going to take a look at it and we'll come back to you. 

Have a wonderful weekend!
Nico.

BennySitbon:
Just an FYI how we found it - we tried directing split to use a specific redis db, lets say `<redis_url>/4`, and we saw it keeps using the `<redis_url>/0` db.
We forked this project, did the above changes, and saw it started using the `/4` db.

NicoZelaya:
@BennySitbon thank you so much for your submission. We're including the PR on our next release.

Apologies for your inconveniences and have a great day! 
Nico.

PR #246, opened 2019-02-23
[WIP] Fix Rubocop offenses

ezesplit:
Removes corresponding classes from exclusions
Creating PR to keep track of changes


lib/splitclient-rb/engine/matchers/matcher.rb
```diff
@@ -0,0 +1,30 @@
...
+  #
+  class Matcher
+    #
+    # evaluates if the given object equals the matcher
+    #
+    # @param obj [object] object to be evaluated
+    #
+    # @return [boolean] true if obj equals the matcher
+    def equals?(obj)
+      if obj.nil?
```
lucas-aragno: this will be handled by `!obj.instance_of?(self.class)` since `nil` is an instance of `NilClass`

lib/splitclient-rb/engine/matchers/matcher.rb
```diff
@@ -0,0 +1,30 @@
...
+        false
+      elsif !obj.instance_of?(self.class)
+        false
+      elsif equal?(obj)
```
lucas-aragno: this makes the rest of the method useless since `equal?` will test if `obj` and `self` are the same if they are the same they won't be true for the other cases, and if they are not equal (by being nil, a different class, or something else) they will go to the `else` case anyway

PR #253, opened 2019-05-31
Event Properties

ezesplit:


lib/splitclient-rb/clients/split_client.rb
```diff
@@ -164,6 +178,33 @@ def sanitize_split_names(calling_method, split_names)
 
     private
 
+    def fix_properties(properties)
```
mmelograno: could you please change the name of the function to something like `validate_properties`? thanks!

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -164,6 +178,33 @@ def sanitize_split_names(calling_method, split_names)
...
+      properties_count = 0
+      non_string_keys = false
+
+      fixed_properties = properties.each_with_object({}) { |(key, value), result|
```
mmelograno: are you doing a loop for checking and then another for calculating the size? why don't add the logic on this `each` method instead of doing two loops?

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -279,5 +320,19 @@ def treatment(
 
       parsed_treatment(multiple, treatment_data)
     end
+
+    def properties_size(properties)
+      properties.inject(0) { |accum, (key, value)|
+        case value
+          when Numeric then accum += value.to_s.length
```
mmelograno: for consistency between the other sdks and simplicity, we are only considering the length of the string

lib/splitclient-rb/cache/repositories/events_repository.rb
```diff
@@ -25,15 +33,23 @@ def metadata
           }
         end
 
-        def event(key, traffic_type, event_type, time, value)
+        def event(key, traffic_type, event_type, time, value, properties, size = nil)
```
mredolatti: Can we not pollute the Event DTO with the size? As an alternative, on other languages we used a wrapper for this.

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -1,4 +1,6 @@
 module SplitIoClient
+  PROPERTIES_SIZE_THRESHOLD = 32768
```
mredolatti: 32kb is the maximum event size. not payload size. We consider each event to have a starting point of 1024 bytes, based on a worst case avg, properties + 1024 should be less than 5mb

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -164,6 +178,33 @@ def sanitize_split_names(calling_method, split_names)
...
+            SplitIoClient.configuration.logger.warn("Property #{key} is of invalid type. Setting value to nil")
+            result[key] = nil
+          end
+        else
+          non_string_keys = true
+          next
+        end
+      }
+
+      SplitIoClient.configuration.logger.warn('Some keys in the properties object are not string. Truncating') if non_string_keys
```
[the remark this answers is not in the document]
ignaciojonas: removed

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -164,6 +178,33 @@ def sanitize_split_names(calling_method, split_names)
...
+        if(key.is_a?(String) || key.is_a?(Symbol))
+          properties_count += 1
+          if value.is_a?(String) || value.is_a?(Symbol) || value.is_a?(Numeric) || value.is_a?(TrueClass) || value.is_a?(FalseClass) || value.nil?
+            result[key] = value
+          else
+            SplitIoClient.configuration.logger.warn("Property #{key} is of invalid type. Setting value to nil")
+            result[key] = nil
+          end
+        else
+          non_string_keys = true
```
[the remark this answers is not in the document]
ignaciojonas: removed

lib/splitclient-rb/clients/split_client.rb
ignaciojonas: changed

lib/splitclient-rb/clients/split_client.rb
ignaciojonas: updated to a single loop

lib/splitclient-rb/clients/split_client.rb
ignaciojonas: Updated

lib/splitclient-rb/cache/repositories/events_repository.rb
ignaciojonas: Removed

lib/splitclient-rb/clients/split_client.rb
ignaciojonas: Changed to `EVENTS_SIZE_THRESHOLD`

PR #256, opened 2019-07-23
Fix latency metrics

ezesplit:


lib/splitclient-rb/cache/repositories/metrics/redis_repository.rb
```diff
@@ -80,6 +78,33 @@ def clear_latencies
             @adapter.delete(keys)
           end
 
+          # introduced to fix incorrect latencies
+          def fix_latencies
+            keys =[]
+
+            22.times do |index|
```
sanzmauro: We need indexs for 0 to 22, I think that this should be `23.times`.

ezesplit: good catch. Changed

PR #258, opened 2019-08-01
Fix logger and validator

ignaciosplit:
* Removed configuration from matchers, just passed the logger.
* Created a logger to manage `log_if_debug` and `log_if_transport`.
* Use a common validator in the configuration to avoid creating multiple instances.


ezesplit:
Closing, and cherry picking approved changes on top of #254 as source branch history has been rewritten

PR #268, opened 2019-09-05
Integrated Localhost Mode

ezesplit:


lib/splitclient-rb/cache/stores/localhost_split_store.rb
```diff
@@ -0,0 +1,193 @@
...
+
+          @splits_repository.add_split(split)
+        end
+
+        def load_features
+          yaml_extensions = [".yml", ".yaml"]
+          if yaml_extensions.include? File.extname(@config.split_file)
+            parse_yaml_features
+          else
+            parse_plain_text_features
```
sanzmauro: I think that we can put the `warn` here and remove the `using_default_split_file` method from the config.  
Putting the log here, we cover all scenarios when the customers use `.split` files.

ezesplit: Done! good catch

PR #280, opened 2019-11-12
Metrics memory repository is triggering a exception

sanzmauro:
# Ruby SDK

## What did you accomplish?
Metrics memory repository is triggering a exception when have `nil` inside the latencies list.

## Extra Notes
Exception message:
```
Unexpected exception in treatment: #<NoMethodError: undefined method `[]' for nil:NilClass> undefined method `[]' for nil:NilClass
```

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,12 +1,12 @@
 7.0.1 (Oct 31, 2019)
  - Updated localhost mode so that parsing of test files results in JSON data for splits analogous to that returned by the Split backend.
- - Resulting localhost data is now used indifferently from actual backend data by the SDK code.
+ - Updated localhost mode to parse the mock data into Split objects to keep differences in operation to a minimum.
  - Removed specific spare code dealing with localhost operations.
  - Removed the disable_impressions configuration param.
  - Fixed an issue about Event properties were not sent to be.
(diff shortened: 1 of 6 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,8 @@
  - Removed the disable_impressions configuration param.
  - Fixed an issue about Event properties were not sent to be.
  - Added ip_addresses_enabled to enable/disable sending MachineName and MachineIP headers when data is posted to Split Servers.
```
NicoZelaya: Try to keep these first words in the past (we've not changed old items, but we should apply the new ways of the changelog) 
Ideally you would use the same four/five words to start the sentences, e.g.: Added, Removed, Updated, Fixed.

To give you an example: 
>  Resulting localhost data is now used indifferently from actual backend data by the SDK code.
could be
> Updated localhost mode to parse the mock data into Split objects to keep differences in operation to a minimum.

CHANGES.txt
```diff
@@ -5,6 +5,8 @@
...
+ - Fixed an issue about attributes in matchers.
+ - Fixed an issue about find latencies.
```
NicoZelaya: The "find latencies" does not mean much externally. Maybe describe the impact? Fixed an issue about trying to access a null property when sending telemetry metrics.

sanzmauro:
@NicoZelaya suggestions applied. Thanks!

PR #288, opened 2019-12-06
Integration tests second pass

sanzmauro:
# Ruby SDK

## What did you accomplish?
- Added integration tests with multiple factories 
- Fixed latencies issue in jruby using Concurrent::Array
- Updated how we create latencies.cleaned redis key removing machine_ip

### DIFF
--- a/spec/integrations/in_memory_client_spec.rb
+++ b/spec/integrations/in_memory_client_spec.rb
@@ -148,6 +148,37 @@ describe SplitIoClient do
       expect(impressions[0][:i][:r]).to eq('not ready')
       expect(impressions[0][:i][:c]).to eq(nil)
     end
+
+    it 'with multiple factories returns on' do
+      local_log = StringIO.new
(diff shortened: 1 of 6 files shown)


spec/integrations/in_memory_client_spec.rb
```diff
@@ -148,6 +148,37 @@
...
+      impressions3 = client3.instance_variable_get(:@impressions_repository).batch
+      impressions4 = client4.instance_variable_get(:@impressions_repository).batch
+
+      expect(impressions1.size).to eq 1
+      expect(impressions2.size).to eq 1
+      expect(impressions3.size).to eq 1
+      expect(impressions4.size).to eq 1
+
+      expect(local_log.string)
+        .to include 'Factory instantiation: You already have 1 factories with this API Key.'
```
NicoZelaya: Ideally we could check how many times this was logged.

sanzmauro: @NicoZelaya suggestions applied.

PR #354, opened 2021-02-17
Fixed missing segment fetch after an SPLIT_UPDATE

sanzmauro:
# Ruby SDK

## What did you accomplish?
- Fixed missing segment fetch after an SPLIT_UPDATE
- Added workers retries.


mredolatti (approved):
Nice work!

PR #413, opened 2022-03-04
Fix typo in SyncManager startup message

geoffharcourt:
There was a typo in a log message that appears when the client boots.

# Ruby SDK

## What did you accomplish?

We fixed the typo!

## How to test new changes?

Boot the client. You're not "streaming" instead of "straming"


agustinona:
Hi Geoff,

Thanks for the catch! Our engineering team will consider this for the next release.

PR #439, opened 2022-10-03
Enabling none impression

sanzmauro:
# Ruby SDK

## What did you accomplish?
- Enabling nene impression

### DIFF
--- a/lib/splitclient-rb/split_config.rb
+++ b/lib/splitclient-rb/split_config.rb
@@ -311,31 +311,17 @@ module SplitIoClient
     end
 
     def init_impressions_mode(impressions_mode, adapter)
-      if adapter == :redis
-        impressions_mode ||= :debug
-
-        case impressions_mode
-        when :optimized
-          return :optimized
-        when :none
-         return :none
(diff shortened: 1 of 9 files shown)


lib/splitclient-rb/split_config.rb
```diff
@@ -310,20 +310,29 @@ def self.default_on_demand_fetch_max_retries
...
+      if adapter == :redis
+        impressions_mode ||= :debug
+
+        case impressions_mode
+        when :optimized
+          return :optimized
+        when :none
+         return :none
+        else
+          @logger.error('You passed an invalid impressions_mode, impressions_mode should be one of the following values: :debug or :optimized. Defaulting to :optimized mode') unless impressions_mode == :optimized
```
gthea, line 323:
```suggestion
          @logger.error('You passed an invalid impressions_mode, impressions_mode should be one of the following values: :debug, :optimized or :none. Defaulting to :optimized mode') unless impressions_mode == :optimized
```

gthea (approved):

PR #473, opened 2023-05-31
added iff compression and decoding message

chillaq:
# Ruby SDK


sanzmauro (changes requested):

lib/splitclient-rb/sse/notification_processor.rb
```diff
@@ -26,7 +26,7 @@ def process(incoming_notification)
 
       def process_split_update(notification)
         @config.logger.debug("SPLIT UPDATE notification received: #{notification}") if @config.debug_enabled
-        @splits_worker.add_to_queue(notification.data['changeNumber'])
+        @splits_worker.split_update(notification)
```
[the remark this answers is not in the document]
chillaq: The adding to the queue happen in splits_worker class, since if it is a IFF there is no need to add it to the queue.

lib/splitclient-rb/sse/workers/splits_worker.rb
```diff
@@ -32,6 +32,20 @@ def stop
           SplitIoClient::Helpers::ThreadHelper.stop(:split_update_worker, @config)
         end
 
+        def split_update(notification)
```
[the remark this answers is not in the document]
chillaq: I am following the same logic as kill_split, it is not called from perform, it does processing, then add fetching the split into the queue.
Why we need to have different logic for splits_update?

lib/splitclient-rb/sse/workers/splits_worker.rb
```diff
@@ -47,6 +61,18 @@ def kill_split(change_number, split_name, default_treatment)
 
         private
 
+        def get_encoded_definition(notification)
```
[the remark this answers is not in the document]
chillaq: sure, will do

lib/splitclient-rb/sse/workers/splits_worker.rb
```diff
@@ -32,25 +32,44 @@ def stop
           SplitIoClient::Helpers::ThreadHelper.stop(:split_update_worker, @config)
         end
 
-        def add_to_queue(change_number)
-          @config.logger.debug("feature_flags_worker add to queue #{change_number}")
-          @queue.push(change_number)
+        def split_update(notification)
```
sanzmauro, line 35: you should check `return if @splits_repository.get_change_number.to_i > notification.data['changeNumber']`

lib/splitclient-rb/sse/workers/splits_worker.rb
```diff
@@ -32,25 +32,44 @@ def stop
...
+          if @splits_repository.get_change_number() == notification.data['pcn']
```
sanzmauro, line 36: ideally you should check if you have correct data in `notification.data['d']`

lib/splitclient-rb/sse/workers/splits_worker.rb
```diff
@@ -32,25 +32,44 @@ def stop
...
+            begin
+              @new_split = JSON.parse(SplitIoClient::Helpers::DecryptionHelper.get_encoded_definition(notification.data['c'], notification.data['d']), symbolize_names: true)
```
sanzmauro, line 38: When you declare a variable with `@` prefix you are declaring an instant variable and is available to all methods within the class. 
In this case you should to declare without `@` given that you need a local variable.

sanzmauro:
I pull your changes and i am getting the following error: 
```
➜  ruby-client git:(iff-update-splitworker) bundle exec rspec

An error occurred while loading ./spec/sse/workers/splits_worker_spec.rb. - Did you mean?
                    rspec ./spec/sse/workers/segments_worker_spec.rb

Failure/Error: require 'byebug'
```

You are missing to add the development dependency in `.gemspec`

lib/splitclient-rb/sse/workers/splits_worker.rb
chillaq: will do

lib/splitclient-rb/sse/workers/splits_worker.rb
chillaq: will remove it

lib/splitclient-rb/sse/workers/splits_worker.rb
chillaq: since it is inside an exception block, any issue with the data['d'] like encoding or compressing, will throw an exception.

sanzmauro: and do you have tests for this scenario ??

chillaq: Yep, in the integration PR

sanzmauro (approved):

PR #482, opened 2023-10-19
Flagset cache flagsetfilter

chillaq:
# Ruby SDK


lib/splitclient-rb/cache/filter/flag_set_filter.rb
```diff
@@ -0,0 +1,38 @@
...
+      class FlagSetsFilter
+        def initialize(flag_sets = [])
+          @flag_sets = Set.new(flag_sets)
+          @should_filter = @flag_sets.any?
+        end
+
+        def flag_set_exist?(flag_set)
+          if not @should_filter
+            return true
+          end
```
sanzmauro, line 17: you can replace this `if` for this line: ` return true unless @should_filter`

lib/splitclient-rb/cache/filter/flag_set_filter.rb
```diff
@@ -0,0 +1,38 @@
...
+          if not flag_set.respond_to?(:lstrip!) or flag_set.empty?
```
sanzmauro, line 18: why you are using `respond_to?` ??

lib/splitclient-rb/cache/filter/flag_set_filter.rb
```diff
@@ -0,0 +1,38 @@
...
+            return false
+          end
+
+          @flag_sets.intersection([flag_set]).any?
+        end
+
+        def intersect?(flag_sets)
+          if not @should_filter
+            return true
+          end
```
sanzmauro, line 28: same here

lib/splitclient-rb/cache/filter/flag_set_filter.rb
chillaq: will do

lib/splitclient-rb/cache/filter/flag_set_filter.rb
chillaq: will do

PR #484, opened 2023-10-23
added flag_set_filter config param and validation

chillaq:
# Ruby SDK

### DIFF
--- a/lib/splitclient-rb/validators.rb
+++ b/lib/splitclient-rb/validators.rb
@@ -4,7 +4,7 @@ require 'set'
 module SplitIoClient
   class Validators
 
-    Flagset_regex = /^[a-zA-Z0-9][-_.:a-zA-Z0-9]{0,79}$/
+    Flagset_regex = /^[a-z0-9][_a-z0-9]{0,49}$/
 
     def initialize(config)
       @config = config
@@ -42,9 +42,8 @@ module SplitIoClient
     end
 
     def valid_flag_sets(method, flag_sets)
-      if flag_sets.nil? || !flag_sets.is_a?(Array)
-        return nil
-      end
+      return Set[] if flag_sets.nil? || !flag_sets.is_a?(Array)
+
       valid_flag_sets = SortedSet[]
       flag_sets.compact.uniq.select do |flag_set|
         if !flag_set.is_a?(String)
@@ -57,7 +56,7 @@ module SplitIoClient
           @config.logger.warn("#{method}: you passed an invalid flag set, flag set name must be a non-empty String")
         end
       end
-      !valid_flag_sets.empty? ? valid_flag_sets :  nil
+      !valid_flag_sets.empty? ? valid_flag_sets :  Set[]
     end
 
     private
--- a/lib/splitclient-rb/validators.rb
+++ b/lib/splitclient-rb/validators.rb
@@ -1,7 +1,10 @@
 # frozen_string_literal: true
+require 'set'
 
 module SplitIoClient
   class Validators
+
+    Flagset_regex = /^[a-zA-Z0-9][-_.:a-zA-Z0-9]{0,79}$/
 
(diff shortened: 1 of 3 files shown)


lib/splitclient-rb/validators.rb
```diff
@@ -1,8 +1,11 @@
 # frozen_string_literal: true
+require 'set'
 
 module SplitIoClient
   class Validators
 
+    Flagset_regex = /^[a-zA-Z0-9][-_.:a-zA-Z0-9]{0,79}$/
```
sanzmauro, line 7: where did this regex come from? 😱 
it should be `"^[a-z0-9][_a-z0-9]{0,49}$"`

lib/splitclient-rb/validators.rb
```diff
@@ -38,6 +41,25 @@ def valid_matcher_arguments(args)
       true
     end
 
+    def valid_flag_sets(method, flag_sets)
```
sanzmauro, line 44: please use the same name defined in the spec. Maybe you can create a `FlagSetsValidator` file with this validations

lib/splitclient-rb/validators.rb
```diff
@@ -38,6 +41,25 @@ def valid_matcher_arguments(args)
...
+          log_invalid_type(:flag_set, method)
+        elsif flag_set.is_a?(String) && flag_set.empty?
+          log_nil(:flag_set, method)
+        elsif !flag_set.empty? && string_match?(flag_set.strip, method)
+          valid_flag_sets.add(flag_set.strip.downcase)
+        else
+          @config.logger.warn("#{method}: you passed an invalid flag set, flag set name must be a non-empty String")
+        end
+      end
+      !valid_flag_sets.empty? ? valid_flag_sets :  nil
```
sanzmauro, line 60: you should return a empty `set` and not `nil` to avoid future failures.

lib/splitclient-rb/validators.rb
```diff
@@ -38,6 +41,25 @@ def valid_matcher_arguments(args)
       true
     end
 
+    def valid_flag_sets(method, flag_sets)
+      if flag_sets.nil? || !flag_sets.is_a?(Array)
+        return nil
+      end
+      valid_flag_sets = SortedSet[]
+      flag_sets.compact.uniq.select do |flag_set|
```
sanzmauro, line 49: you are not trimming the `flag set` if neecessary.

lib/splitclient-rb/validators.rb
```diff
@@ -38,6 +41,25 @@ def valid_matcher_arguments(args)
       true
     end
 
+    def valid_flag_sets(method, flag_sets)
+      if flag_sets.nil? || !flag_sets.is_a?(Array)
```
sanzmauro, line 45: you can replace this if for one line: `return nil if flag_sets.nil? || !flag_sets.is_a?(Array)`

lib/splitclient-rb/validators.rb
chillaq: Hmmm.. this is used in Python SDK, I will correct both.

lib/splitclient-rb/validators.rb
chillaq: will do

lib/splitclient-rb/validators.rb
chillaq: The trimming is done inside the loop, I need to check first if the object is string, otherwise the strip method will raise exception

lib/splitclient-rb/validators.rb
chillaq: will do

lib/splitclient-rb/validators.rb
chillaq: We agreed to make it technical debt

sanzmauro (approved):

PR #493, opened 2023-11-10
Flagset evaluator refactor

chillaq:
# Ruby SDK

### DIFF
--- a/lib/splitclient-rb/clients/split_client.rb
+++ b/lib/splitclient-rb/clients/split_client.rb
@@ -29,16 +29,15 @@ module SplitIoClient
       @config = config
       @impressions_manager = impressions_manager
       @telemetry_evaluation_producer = telemetry_evaluation_producer
-      @split_validator = SplitIoClient::Validators.new(self)
+      @split_validator = SplitIoClient::Validators.new(@config)
+      @evaluator = Engine::Parser::Evaluator.new(@segments_repository, @splits_repository, @config)
     end
 
     def get_treatment(
         key, split_name, attributes = {}, split_data = nil, store_impressions = true,
         multiple = false, evaluator = nil
       )
-      impressions = []
-      result = treatment(key, split_name, attributes, split_data, store_impressions, multiple, evaluator, GET_TREATMENT, impressions)
-      @impressions_manager.track(impressions)
+      result = treatment(key, split_name, attributes, split_data, store_impressions, multiple, GET_TREATMENT)
 
       if multiple
         result.tap { |t| t.delete(:config) }
@@ -51,15 +50,12 @@ module SplitIoClient
         key, split_name, attributes = {}, split_data = nil, store_impressions = true,
         multiple = false, evaluator = nil
       )
-      impressions = []
-      result = treatment(key, split_name, attributes, split_data, store_impressions, multiple, evaluator, GET_TREATMENT_WITH_CONFIG, impressions)
-      @impressions_manager.track(impressions)
-
-      result
+      treatment(key, split_name, attributes, split_data, store_impressions, multiple, GET_TREATMENT_WITH_CONFIG)
     end
 
     def get_treatments(key, split_names, attributes = {})
       treatments = treatments(key, split_names, attributes)
+
       return treatments if treatments.nil?
       keys = treatments.keys
       treats = treatments.map { |_,t| t[:treatment]  }
@@ -180,6 +176,9 @@ module SplitIoClient
     end
 
     def sanitize_split_names(calling_method, split_names)
+      if split_names.nil?
+        return nil
+      end
       split_names.compact.uniq.select do |split_name|
         if (split_name.is_a?(String) || split_name.is_a?(Symbol)) && !split_name.empty?
           true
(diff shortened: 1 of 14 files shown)


lib/splitclient-rb/clients/split_client.rb
```diff
@@ -245,25 +247,54 @@ def treatments(key, split_names, attributes = {}, calling_method = 'get_treatmen
...
-      split_names_keys = treatments_labels_change_numbers.keys
-      treatments = treatments_labels_change_numbers.values.map do |v|
+      valid_feature_flag_names = []
+      sanitized_feature_flag_names.each { |feature_flag_name|
+        valid_feature_flag_names << feature_flag_name unless feature_flag_name.nil?
+      }
+      start = Time.now
+      impressions_total = []
+
+      feature_flags = @splits_repository.splits(feature_flag_names)
```
sanzmauro, line 276: why are you `splits_repository` here? also you should use `Evaluator.evaluate_feature_flags` in this method

lib/splitclient-rb/engine/parser/evaluator.rb
```diff
@@ -2,41 +2,60 @@ module SplitIoClient
...
+        def initialize(segments_repository, splits_repository, config)
           @splits_repository = splits_repository
           @segments_repository = segments_repository
-          @multiple = multiple
           @config = config
           @cache = {}
         end
 
-        def call(keys, split, attributes = nil)
+        def evaluate_feature_flag(keys, feature_flag, attributes = nil)
```
sanzmauro, line 12: i think that you should get the `Feature Flag definition` here and not in the client to be consistency between methods

spec/engine/parser/evaluator_spec.rb
```diff
@@ -4,7 +4,9 @@
 
 describe SplitIoClient::Engine::Parser::Evaluator do
```
sanzmauro, line 5: please add coverage for `evaluator.evaluate_feature_flags`

sanzmauro (changes requested):

lib/splitclient-rb/engine/parser/evaluator.rb
```diff
@@ -2,41 +2,60 @@ module SplitIoClient
...
-          treatment = if Models::Split.matchable?(split)
-            if @multiple && @cache[digest]
-              @cache[digest]
-            else
-              match(split, keys, attributes)
-            end
+        private
+
+        def evaluate_treatment(keys, feature_flag, cache_result, attributes)
+          digest = Digest::MD5.hexdigest("#{{matching_key: keys[:matching_key]}}#{feature_flag}#{attributes}")
```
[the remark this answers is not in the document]
chillaq: This is part of the original code I kept

lib/splitclient-rb/engine/parser/evaluator.rb
```diff
@@ -2,41 +2,60 @@ module SplitIoClient
...
+          end
+
+          treatment = if Models::Split.matchable?(feature_flag)
+            match(feature_flag, keys, attributes)
           else
-            treatment_hash(Models::Label::KILLED, split[:defaultTreatment], split[:changeNumber], split_configurations(split[:defaultTreatment], split))
+            treatment_hash(Models::Label::KILLED, feature_flag[:defaultTreatment], feature_flag[:changeNumber], split_configurations(feature_flag[:defaultTreatment], feature_flag))
           end
 
           @cache[digest] = treatment if cache_result
```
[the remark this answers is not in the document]
chillaq: Yes, it will be return its value to "treatment" variable when the if condition is false

lib/splitclient-rb/clients/split_client.rb
chillaq: I did in the beginning, but it became too complicated to cover all scenarios of result treatment, it was cleaner to fetch splits in client then evaluate each single split.

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -29,37 +29,28 @@ def initialize(sdk_key, repositories, status_manager, config, impressions_manage
       @config = config
       @impressions_manager = impressions_manager
       @telemetry_evaluation_producer = telemetry_evaluation_producer
-      @split_validator = SplitIoClient::Validators.new(self)
+      @split_validator = SplitIoClient::Validators.new(@config)
+      @evaluator = Engine::Parser::Evaluator.new(@segments_repository, @splits_repository, @config)
```
sanzmauro, line 33: we can move this to the factory level.

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -29,37 +29,28 @@ def initialize(sdk_key, repositories, status_manager, config, impressions_manage
       @config = config
       @impressions_manager = impressions_manager
       @telemetry_evaluation_producer = telemetry_evaluation_producer
-      @split_validator = SplitIoClient::Validators.new(self)
+      @split_validator = SplitIoClient::Validators.new(@config)
```
sanzmauro, line 32: same

lib/splitclient-rb/clients/split_client.rb
```diff
@@ -163,23 +160,17 @@ def keys_from_key(key)
...
-      else
-        {
-          treatment: treatment_data[:treatment],
-          config: treatment_data[:config],
-        }
-      end
     end
 
     def sanitize_split_names(calling_method, split_names)
+      if split_names.nil?
```
sanzmauro, line 171: `return nil if split_names.nil?`

sanzmauro (changes requested):
are you running `rubocop` before pushing? I am getting `10 offenses detected and  6 offenses auto-correctable`

lib/splitclient-rb/clients/split_client.rb
chillaq: will do

sanzmauro (approved):

PR #500, opened 2023-12-04
Added sets to manager split view dict

chillaq:
# Ruby SDK


lib/splitclient-rb/managers/split_manager.rb
```diff
@@ -105,7 +105,8 @@ def build_split_view(name, split)
           killed: split[:killed],
           treatments: treatments,
           change_number: split[:changeNumber],
-          configs: split[:configurations] || {}
+          configs: split[:configurations] || {},
+          sets: split[:sets]
```
sanzmauro, line 109: if sets is `nil` should return `{}`

sanzmauro (approved):

sanzmauro (approved):

PR #508, opened 2023-12-19
Jsonlh engine api split

chillaq:
# Ruby SDK


lib/splitclient-rb/engine/api/json_localhost/splits.rb
```diff
@@ -0,0 +1,160 @@
+# frozen_string_literal: true
+
+module SplitIoClient
+  module Api
+    # Retrieves split definitions from the Split Backend
```
sanzmauro, line 5: 🤔 🤔

lib/splitclient-rb/engine/api/json_localhost/splits.rb
```diff
@@ -0,0 +1,160 @@
...
+          parsed = JSON.parse(File.read(@split_file), symbolize_names: true)
+          santitized = sanitize_feature_flag(parsed)
+          return santitized
+        rescue StandardError => e
+          @config.logger.error("Exception caught: #{e.message}")
+          raise "Error parsing splits file \'#{@split_file}\', Make sure it's readable."
+        end
+      end
+
+      def sanitize_feature_flag(parsed)
```
sanzmauro, line 58: maybe you can create a `splits_sanitizer.rb` and move all this methods

lib/splitclient-rb/engine/api/json_localhost/splits.rb
chillaq: This class is used in the engine in the same way the Api classes used (for http and redis synching), that is why I added it here.

lib/splitclient-rb/engine/api/json_localhost/splits.rb
chillaq: Sure, will do.

sanzmauro (approved):

PR #514, opened 2024-03-26
Semver class

chillaq:
# Ruby SDK

### DIFF
--- a/spec/test_data/splits/semver/between-semver.csv
+++ b/spec/test_data/splits/semver/between-semver.csv
@@ -1,0 +1,18 @@
+# version1, version2, version3, expected
+1.1.1,2.2.2,3.3.3,true
+1.1.1-rc.1,1.1.1-rc.2,1.1.1-rc.3,true
+1.0.0-alpha,1.0.0-alpha.1,1.0.0-alpha.beta,true
+1.0.0-alpha.1,1.0.0-alpha.beta,1.0.0-beta,true
+1.0.0-alpha.beta,1.0.0-beta,1.0.0-beta.2,true
+1.0.0-beta,1.0.0-beta.2,1.0.0-beta.11,true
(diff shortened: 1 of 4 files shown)


spec/engine/matchers/semver_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'spec_helper'
+
+describe SplitIoClient::Semver do
+  let(:valid_versions) { ["1.1.2", "1.1.1", "1.0.0", "1.0.0-rc.1", "1.0.0-beta.11", "1.0.0-beta.2",
```
sanzmauro, line 4: Why are you not using the csv example? 
The main idea of the csv file is to have the same tests across the SDKs, and if we want to update the file or add more examples, the changes are easier.

lib/splitclient-rb/engine/matchers/semver.rb
```diff
@@ -0,0 +1,171 @@
...
+    #
+    # @param version [String] raw version as read from splitChanges response.
+    def initialize(version)
+      @major = 0
+      @minor = 0
+      @patch = 0
+      @pre_release = []
+      @is_stable = false
+      @old_version = version
+      parse
```
sanzmauro, line 22: please don't parse at constructor level.
if you have a `RuntimeError` at this level you are not catching it.

sanzmauro (changes requested):

spec/engine/matchers/semver_spec.rb
chillaq: will do

lib/splitclient-rb/engine/matchers/semver.rb
chillaq: will do

sanzmauro (approved):

PR #530, opened 2024-04-24
added spec query parameter

chillaq:
# Ruby SDK

### DIFF
--- a/lib/splitclient-rb/engine/api/splits.rb
+++ b/lib/splitclient-rb/engine/api/splits.rb
@@ -4,6 +4,9 @@ module SplitIoClient
   module Api
     # Retrieves split definitions from the Split Backend
     class Splits < Client
+
+      SPEC_VERSION = "1.1"
+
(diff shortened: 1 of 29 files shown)


lib/splitclient-rb/engine/api/splits.rb
```diff
@@ -4,17 +4,20 @@ module SplitIoClient
   module Api
     # Retrieves split definitions from the Split Backend
     class Splits < Client
+
+      SPEC_VERSION = "1.1"
```
sanzmauro, line 8: move this definition into a separate file in the root of the project please 🙏

```
For consistency across our SDKs and to streamline reviews, issue tracking, and support, we aim to include a new file named spec at the root of each project. This file will define and allow us to update the value of the feature flags spec.
```

chillaq: will do

sanzmauro (approved):

PR #552, opened 2025-05-15
Added old spec support

chillaq:
# Ruby SDK


spec/sse/workers/splits_worker_spec.rb
```diff
@@ -309,7 +309,7 @@
       "trafficTypeName": "user",
       "excluded":{
         "keys":["mauro@split.io","gaston@split.io"],
-        "segments":[]
+        "segments":["segment1"]
```
EmilianoSanchez, line 312: Update to `"segments":[{ "type": "standard", "name": "segment1"}]` if not done in a different PR.

EmilianoSanchez (approved):

spec/sse/workers/splits_worker_spec.rb
chillaq: will do

PR #564, opened 2025-07-02
Updated engine and sync classes

chillaq:
# Ruby SDK

### DIFF
--- a/spec/cache/senders/impressions_formatter_spec.rb
+++ b/spec/cache/senders/impressions_formatter_spec.rb
@@ -58,7 +58,8 @@ describe SplitIoClient::Cache::Senders::ImpressionsFormatter do
                                 b: 'foo1',
                                 r: 'custom_label1',
                                 c: 123_456,
-                                pt: nil }]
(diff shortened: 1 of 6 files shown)


spec/cache/senders/impressions_formatter_spec.rb
```diff
@@ -68,14 +69,15 @@
                                  b: 'foo2',
                                  r: 'custom_label2',
                                  c: 123_499,
-                                 pt: nil }]
+                                pt: nil,
+                                properties: "null" }]
```
EmilianoSanchez, line 73: if there are no properties, shouldn't it be `properties: nil`? and so, no `properties` field is sent in the `POST /testImpressions/bulk` payload.

EmilianoSanchez (changes requested):

spec/cache/senders/impressions_formatter_spec.rb
chillaq: Correct, will fix it.

PR #566, opened 2025-07-07
Updated integration tests

chillaq:
# Ruby SDK


lib/splitclient-rb/clients/split_client.rb
```diff
@@ -273,6 +273,7 @@ def treatments(key, feature_flag_names, attributes = {}, options = nil, calling_
       bucketing_key, matching_key = keys_from_key(key)
       bucketing_key = bucketing_key ? bucketing_key.to_s : nil
       matching_key = matching_key ? matching_key.to_s : nil
+      attributes = parsed_attributes(attributes)
```
sanzmauro, line 276: why are you adding this?

PR #625, opened 2026-06-15
fix: add parentheses to is_a? call in SSEClient#close

### LINKED ISSUE #623: TypeError in SSEClient#close from paren-less is_a? call (8.11.0)
## Description

Version 8.11.0 introduced a `TypeError: class or module required` in `SSEClient#close` that fires on every SSE socket close. The error is caught by the `rescue StandardError` block so it doesn't crash the process, but it generates a high volume of error log entries (~400/day in our production environment across multiple services consuming the SDK).

We had to pin `splitclient-rb < 8.11.0` in our wrapper gem to avoid this.

## SDK Version

- **Affected:** 8.11.0, master
- **Not affected:** 8.10.1 and earlier

## Root Cause

In `lib/splitclient-rb/sse/event_source/client.rb`, line 50 of the `close` method:

```ruby
@config.logger.debug("SSEClient socket state #{@socket.state}") if @socket.is_a? OpenSSL::SSL::SSLSocket && @config.debug_enabled
```

Without explicit parentheses, Ruby parses the `is_a?` argument as:

```ruby
@socket.is_a?(OpenSSL::SSL::SSLSocket && @config.debug_enabled)
```

Since `OpenSSL::SSL::SSLSocket` is truthy, `&&` evaluates and returns the right operand (`@config.debug_enabled`, a boolean). Then `is_a?(true)` or `is_a?(false)` raises `TypeError: class or module required`.

Note that line 48 has a similar `is_a?` call but without `&&`, so it parses correctly:
```ruby
@socket.sync_close = true if @socket.is_a? OpenSSL::SSL::SSLSocket  # works fine
```

## Suggested Fix

Add explicit parentheses:

```ruby
@config.logger.debug("SSEClient socket state #{@socket.state}") if @socket.is_a?(OpenSSL::SSL::SSLSocket) && @config.debug_enabled
```

Happy to submit a PR for this if that's helpful.

dan98765:
## What did you accomplish?

Fixed a `TypeError: class or module required` in `SSEClient#close` (line 50 of `lib/splitclient-rb/sse/event_source/client.rb`).

Without explicit parentheses, Ruby parses:

```ruby
if @socket.is_a? OpenSSL::SSL::SSLSocket && @config.debug_enabled
```

as:

```ruby
if @socket.is_a?(OpenSSL::SSL::SSLSocket && @config.debug_enabled)
```

Since `OpenSSL::SSL::SSLSocket` is truthy, `&&` evaluates and returns the right operand (`@config.debug_enabled`, a boolean). Then `is_a?(true)` or `is_a?(false)` raises `TypeError`.

The `rescue StandardError` block catches it so it doesn't crash, but it fires on every SSE socket close and produces error-level log entries.

The fix adds explicit parentheses: `@socket.is_a?(OpenSSL::SSL::SSLSocket)`.

Note that line 48 has a similar `is_a?` call without `&&`, so it parses correctly and doesn't need a change.

Fixes #623.

## How to test new changes?

The fix is a Ruby parsing issue, not a logic change. You can verify the parsing difference in an IRB session:

```ruby
obj = OpenSSL::SSL::SSLSocket
debug = true

# Without parens (current code) — TypeError
obj.is_a? OpenSSL::SSL::SSLSocket && debug
# => TypeError: class or module required

# With parens (this fix) — works correctly
obj.is_a?(OpenSSL::SSL::SSLSocket) && debug
# => true
```

## Extra Notes

This was introduced in 8.11.0 when the `&& @config.debug_enabled` guard was added to the `is_a?` check.

### DIFF
--- a/lib/splitclient-rb/engine/matchers/rule_based_segment_matcher.rb
+++ b/lib/splitclient-rb/engine/matchers/rule_based_segment_matcher.rb
@@ -40,7 +40,7 @@ module SplitIoClient
         break if matched
       end
 
-      @logger.debug("[InRuleSegmentMatcher] #{@segment_name} is in rule based segment -> #{matched}")
+      @config.logger.debug("[InRuleSegmentMatcher] #{@segment_name} is in rule based segment -> #{matched}") if @config.debug_enabled
       matched
     end
 
--- a/lib/splitclient-rb/engine/sync_manager.rb
+++ b/lib/splitclient-rb/engine/sync_manager.rb
@@ -6,14 +6,8 @@ module SplitIoClient
       SYNC_MODE_STREAMING = 0
       SYNC_MODE_POLLING = 1
 
-      def initialize(config,
-                     synchronizer,
-                     telemetry_runtime_producer,
-                     telemetry_synchronizer,
-                     status_manager,
-                     sse_handler,
-                     push_manager,
-                     status_queue)
+      def initialize(config, synchronizer, telemetry_runtime_producer, telemetry_synchronizer,
+                     status_manager, sse_handler, push_manager, status_queue)
         @config = config
         @synchronizer = synchronizer
         @telemetry_runtime_producer = telemetry_runtime_producer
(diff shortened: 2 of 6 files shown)


chillaq (approved):

dan98765:
Closing — this is superseded by the fix in 8.11.1. Thanks @chillaq!

